PR #7, opened 2022-01-20
fix: general improvements

wtrocki:
- Moved pages to _pages doc
- Moved to adoc (apart from adr.md that was breaking scripts)
- Created sepratate header (process)
- included cname
- Verified locally


wtrocki:
@tombentley Website is little bit broken right now with links not working and domain will be removed every time we make change. This PR is going to fix it.

tombentley (approved):

PR #9, opened 2022-01-20
More improvements

tombentley:


pages/about/adr-process.adoc
```diff
@@ -0,0 +1,57 @@
...
+If you identify the need for an ADR please {{site.github.repository_url}}/issues/new[open an issue] describing the decision that needs to be taken.
+
+The discussion of the issue should decide whether a new ADR really is needed.
+If so, it should also seek to identify.
+
+* Whether any {{site.url}}/padr/[platform ADRs] apply. 
+* What tags the ADR should have. 
+* What the _summary_ of the new ADR should be. 
+
+=== Drafting an ADR
```
wtrocki, line 21: I can automate that step easily after.
When approved label is added to issue we can trigger bot that will

close issue
Create PR with title and number from issue.

pages/adr.adoc
```diff
@@ -31,72 +32,3 @@ If you don't know what an ADR is, look link:/about/#architecture-decision-record
 </table>
 ++++
 
-== Github Process
-
-Logically an ADR can be in a number of states (Proposed, In Progress, In
-Review, Accepted, Rejected, Deferred or Superseded) with the following
-state machine
-
-....
```
wtrocki, line 40: [optional] I like that diagram, but cannot see it in new structure

wtrocki (approved):

_includes/footer.html
```diff
@@ -1,3 +1,7 @@
+<div class="edit">
+  <a href="{{site.github.repository_url}}/edit/main/{{page.path}}">✎ Edit this page</a>
```
wtrocki, line 2: We got build issue based of that value missing
https://github.com/redhat-developer/app-services-architecture/runs/4887126419?check_suite_focus=true

wtrocki: actually that is other issue. Let me fix that

PR #10, opened 2022-01-20
add github metadata

wtrocki:


wtrocki:
This somehow did not work. Easier to add repo to config

_config.yml
```diff
@@ -43,6 +43,7 @@ plugins:
   - jekyll-titles-from-headings
   - jekyll-relative-links
   - jekyll-avatar
+  - jekyll-github-metadata
```
tombentley, line 46: This was already added on line 39. TBH I don't really understand the difference between adding in the gem and adding in the `_config.yml`. Are both required?

wtrocki: It works locally but not on CI/CD.

wtrocki: As for difference it depends on version of jeckyll but yes. this PR did not made any fix. Sorry

PR #11, opened 2022-01-21
Override github url 

wtrocki:
Reverts redhat-developer/app-services-architecture#10

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -54,6 +54,9 @@ collections:
 
 include:
  - CNAME
+ 
+ github:
+  repository_url: https://github.com/redhat-developer/app-services-architecture
 
 
 # Exclude from processing.
--- a/_config.yml
+++ b/_config.yml
@@ -43,7 +43,6 @@ plugins:
   - jekyll-titles-from-headings
   - jekyll-relative-links
   - jekyll-avatar
-  - jekyll-github-metadata
 
 collections:
   adr:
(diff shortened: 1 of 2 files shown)


wtrocki:
Plugin provides metadata and it works locally but somehow fails on deployment:
https://jekyll.github.io/github-metadata/site.github/

Not sure why I done hack based on docs:
https://github.com/pages-themes/leap-day#overriding-github-generated-urls

So that will specify this value explicitly.

_config.yml
```diff
@@ -55,6 +54,9 @@ collections:
 
 include:
  - CNAME
+ 
+ github:
```
tombentley, line 58: Probably worth a comment as to why we're doing this. E.g. a link to https://github.com/redhat-developer/app-services-architecture/pull/11 would be enough, I think.

tombentley (approved):

wtrocki:
Still not sure if that is going to work. It always worked locally.

tombentley:
Let's try it. I suppose that upgrading jekyll would be an alternative thing to try, but a bigger change than this.

wtrocki:
Actually I have missed that info in logs when locally:

 >  GitHub Metadata: No GitHub API authentication could be found. Some fields may be missing or have incorrect data.
 
 So proper fix is to not use that and hardcode values or use github api key

PR #19, opened 2022-02-08
ADR 80  CEE diagnostic gathering mechanism

k-wall:
In order to provide support to customers of a Managed Service, CEE require a mechanism that allows them to capture the
state of the instance of the service for diagnostic purposes.  CEE will use this information in conjunction with the
problem statement provided by the customer in order to efficiently investigate the problem and determine a solution.

Managed Kafka has adopted backplane managed scripts as a solution to this problem. We want to propose it as a pattern.

### DIFF
--- a/_adr/80/index.adoc
+++ b/_adr/80/index.adoc
@@ -6,13 +6,15 @@ authors:
   - "Keith Wall"
 ---
 
+include::/_includes/glossary_links.adoc[]
+
 ## Context and Problem Statement
 
-In order to provide support to the customers of a Managed Service, a company offering the Managed Service will typically have a Support
-Organisation (SO) dedicated to this function.  In order to perform this function effectively the SO requires a mechanism that allows
-them to capture the state of the instance of the service for diagnostic purposes.
+In order to provide support to the customers of a Managed Service, a company offering the Managed Service will typically have a {glossary-customersupport}
+organisation.  In order to assist the customers, members of {glossary-customersupport} require a mechanism that allows them to capture the state of the
+instance of the service for diagnostic purposes.
 
-The SO will use the gathered information in conjunction with the problem statement provided by the customer in order to efficiently
+{glossary-customersupport} will use the gathered information in conjunction with the problem statement provided by the customer in order to efficiently
 investigate the problem and determine a solution.
 
 The data gathering mechanism must consider the issue of Personally Identifiable Information (PII) or other sensitive information and ensure
@@ -31,10 +33,9 @@ permissioned groups.
 None
 
 ## Stakeholders
-* Eng
-* BU
-* SRE
-* SO
+* {glossary-businessstakeholder}
+* {glossary-sre}
+* {glossary-customersupport}
 
 ## Current Architecture
 
@@ -47,7 +48,7 @@ https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used in thi
 It is RECOMMENDED that managed services use https://source.redhat.com/groups/public/openshiftplatformsre/wiki/backplane_scripts[Back Plane]
 https://github.com/openshift/managed-scripts/[managed scripts] as the foundation of the data gathering mechanism. Backplane has an
 established https://source.redhat.com/groups/public/openshiftplatformsre/wiki/backplane_cluster_permissions[authorization model] that ensures
-that members of the SO have the permissions to run the scripts.
+that members of the {glossary-customersupport} have the permissions to run the scripts.
 
 Each Managed Service that implements this ADR MUST provide one or more backplane scripts that interrogate the state of a service instance.
 The exact way way the script will interrogate state will be service specific, however if the service provided on top of OpenShift, it is
--- a/_adr/80/index.adoc
+++ b/_adr/80/index.adoc
@@ -66,14 +66,12 @@ The scripts MUST be non-mutating in nature.
 
 ### Threat Model
 
-* <Provide a link to the relevant threat model. You must either update an existing threat model(s) to cover the changes made by this ADR, or add a new threat model.
-
-* Take a copy of the threat model when the ADR is proposed, and merge those changes in once the ADR is accepted
+This ADR does not alert the threat model.
 
 ## Alternatives Considered / Rejected
 
-The option of providing an OpenShift MustGather Image was considered, however, as the approved access mechanism to the data-plane is backplane,
-utilising the backplane managed-scripts feature to perform the data gathering step minimised dependencies.
+The option of providing an https://access.redhat.com/articles/4591521[OpenShift MustGather Image] was considered, however, as the approved access
+mechanism to the data-plane is backplane, utilising the backplane managed-scripts feature to perform the data gathering step minimised dependencies.
 
 ## Challenges
 
--- a/_adr/80/index.adoc
+++ b/_adr/80/index.adoc
@@ -46,7 +46,7 @@ It is RECOMMENDED that managed services use https://source.redhat.com/groups/pub
 https://github.com/openshift/managed-scripts/[managed scripts] as the foundation of the  data gathering mechanism. Backplane has an 
 established authorization model that ensures that CEE have the permissions to run the scripts (KW: how does this work)?
 
-Each Managed Service that implements this ADR MUST provide one or more backplane scripts that the interrogate the state of a service instance.
+Each Managed Service that implements this ADR MUST provide one or more backplane scripts that interrogate the state of a service instance.
 The exact way way the script will interrogate state will be service specific, however if the service provided on top of OpenShift, it is
 RECOMMENEDED that the scropt invoke the `oc adm inspect` command to gather than Kubernetes state in addition to any other service specific
 state gathering commands.
--- a/_adr/80/index.adoc
+++ b/_adr/80/index.adoc
@@ -48,7 +48,7 @@ https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used in thi
 It is RECOMMENDED that managed services use https://source.redhat.com/groups/public/openshiftplatformsre/wiki/backplane_scripts[Back Plane]
 https://github.com/openshift/managed-scripts/[managed scripts] as the foundation of the data gathering mechanism. Backplane has an
 established https://source.redhat.com/groups/public/openshiftplatformsre/wiki/backplane_cluster_permissions[authorization model] that ensures
-that members of the {glossary-customersupport} have the permissions to run the scripts.
+that members of {glossary-customersupport} have the permissions to run the scripts.
 
 Each Managed Service that implements this ADR MUST provide one or more backplane scripts that interrogate the state of a service instance.
 The exact way way the script will interrogate state will be service specific, however if the service provided on top of OpenShift, it is
--- a/_adr/80/index.adoc
+++ b/_adr/80/index.adoc
@@ -51,14 +51,14 @@ established https://source.redhat.com/groups/public/openshiftplatformsre/wiki/ba
 that members of {glossary-customersupport} have the permissions to run the scripts.
 
 Each Managed Service that implements this ADR MUST provide one or more backplane scripts that interrogate the state of a service instance.
-The exact way way the script will interrogate state will be service specific, however if the service provided on top of OpenShift, it is
-RECOMMENDED that the script invoke the `oc adm inspect` command to gather Kubernetes state in addition to any other service specific
+The exact way way the scripts will interrogate state will be service specific, however if the service provided on top of OpenShift, it is
+RECOMMENDED that the scripts invoke the `oc adm inspect` command to gather Kubernetes state in addition to any other service specific
 state gathering commands.
 
-It is RECOMMENDED that the output from the backplane script be a UNIX tar stream.  This will allow the artifact to provide the diagnostic
+It is RECOMMENDED that the output from the scripts be a UNIX tar stream.  This will allow the artifact to provide the diagnostic
 information in a structured form that can easily be processed using common command-line tools.
 
-It is RECOMMENDED that the script takes reasonable steps to avoid the collection of PII or other sensitive information. This may be done
+It is RECOMMENDED that the scripts takes reasonable steps to avoid the collection of PII or other sensitive information. This may be done
 either by redacting artifacts known to contain secrets/keys etc, or not gathering certain information in the first place.  For
 example, in a Java based system where PII may be present in the Java heap, the safe option is not to collect heap dumps.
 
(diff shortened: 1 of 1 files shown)


k-wall:
@StevenTobin  would you mind giving this a pre-review before I send out for a wider audience?

_adr/80/index.adoc
```diff
@@ -20,3 +20,73 @@ permissioned groups.
...
+
+None
+
+## Proposed Architecture
+
+https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used in this proposal.
+
+It is RECOMMENDED that managed services use https://source.redhat.com/groups/public/openshiftplatformsre/wiki/backplane_scripts[Back Plane]
+https://github.com/openshift/managed-scripts/[managed scripts] as the foundation of the  data gathering mechanism. Backplane has an 
+established authorization model that ensures that CEE have the permissions to run the scripts (KW: how does this work)?
```
StevenTobin, line 47: This [document](https://source.redhat.com/groups/public/openshiftplatformsre/wiki/backplane_cluster_permissions) describes how the permissions are set up.

RBAC rules are defined [here](https://github.com/openshift/managed-cluster-config/tree/master/deploy/backplane)

_adr/80/index.adoc
```diff
@@ -20,3 +20,73 @@ permissioned groups.
...
+
+Each Managed Service that implements this ADR MUST provide one or more backplane scripts that the interrogate the state of a service instance.
```
StevenTobin, line 49:
```suggestion
Each Managed Service that implements this ADR MUST provide one or more backplane scripts that interrogate the state of a service instance.
```

_adr/80/index.adoc
```diff
@@ -20,3 +20,73 @@ permissioned groups.
...
+The exact way way the script will interrogate state will be service specific, however if the service provided on top of OpenShift, it is
+RECOMMENEDED that the scropt invoke the `oc adm inspect` command to gather than Kubernetes state in addition to any other service specific
```
StevenTobin, line 51:
```suggestion
RECOMMENDED that the script invoke the `oc adm inspect` command to gather Kubernetes state in addition to any other service specific
```

StevenTobin (commented):
Marked some small typos but content LGTM

_adr/80/index.adoc
StevenTobin: I don't think it's necessary to include this information however. We should be able to assume that CEE and SRE can run any scripts provided.

k-wall: thanks, I linked the document you provided to help a reader understand the authz model, but as you suggest, I've included no details in this doc.

_adr/80/index.adoc
```diff
@@ -12,11 +12,80 @@ In order to provide support to customers of a Managed Service, CEE require a mec
 state of the instance of the service for diagnostic purposes.  CEE will use this information in conjunction with the
 problem statement provided by the customer in order to efficiently investigate the problem and determine a solution.
 
-The data gathering mechanism must consider the issue of Personally Identifiable Information (PII) or other sensitive
-information and ensure that it is handled appropriately.
+The data gathering mechanism must consider the issue of
+https://source.redhat.com/departments/it/digitalsolutionsdelivery/enterprise-architecture/mojo_content/red_hat_data_classifications
```
emmanuelbernard, line 16: This is a reference to a private resource, let's avoid that.

_adr/80/index.adoc
```diff
@@ -12,11 +12,80 @@ In order to provide support to customers of a Managed Service, CEE require a mec
 state of the instance of the service for diagnostic purposes.  CEE will use this information in conjunction with the
 problem statement provided by the customer in order to efficiently investigate the problem and determine a solution.
 
-The data gathering mechanism must consider the issue of Personally Identifiable Information (PII) or other sensitive
-information and ensure that it is handled appropriately.
+The data gathering mechanism must consider the issue of
```
emmanuelbernard, line 15: General comment

I think this ADR needs a different title.

Since these are open source, any obscure function name is not helping.
I propose to rename CEE with a support organization.

emmanuelbernard (commented):
This proposed ADR is interesting as it references extensively some Red Hat specific mechanism. I've started a conversation to see how we want to address this.

_adr/80/index.adoc
k-wall: > This proposed ADR is interesting as it references extensively some Red Hat specific mechanism. I've started a conversation to see how we want to address this.

I see at least some of Backplane is open, https://github.com/openshift/managed-scripts, but how complete/usable the open source offering is for someone say running kube  I don't understand the Backplane landscape sufficiently well to answer that.

_adr/80/index.adoc
```diff
@@ -1,22 +1,90 @@
...
+The exact way way the script will interrogate state will be service specific, however if the service provided on top of OpenShift, it is
+RECOMMENDED that the script invoke the `oc adm inspect` command to gather Kubernetes state in addition to any other service specific
+state gathering commands.
+
+It is RECOMMENDED that the output from the backplane script be a UNIX tar stream.  This will allow the artifact to provide the diagnostic
+information in a structured form that can easily be processed using common command-line tools.
+
+The scripts MUST NOT gather PII or other sensitive information.  This may be done either by redacting artifacts known to contain secrets, or
+not gathering certain information in the first place.  For example, in a Java based system where PII may be present in the Java heap, the safe
+option is not to collect heap dumps.
```
tombentley, line 62: This could be impossible to guarantee in practice. E.g. with Kafka how do I know that a service user hasn't created topics whose names include SSN? While in Kafka that would be an unusual pattern, it might be more common in other services for identifiers to leak from data handled by the service into something your scripts should collect.

_adr/80/index.adoc
```diff
@@ -1,22 +1,90 @@
...
+
+None
+
+## Proposed Architecture
+
+https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used in this proposal.
+
+It is RECOMMENDED that managed services use https://source.redhat.com/groups/public/openshiftplatformsre/wiki/backplane_scripts[Back Plane]
+https://github.com/openshift/managed-scripts/[managed scripts] as the foundation of the data gathering mechanism. Backplane has an
+established https://source.redhat.com/groups/public/openshiftplatformsre/wiki/backplane_cluster_permissions[authorization model] that ensures
```
tombentley, line 49: Non-public link

_adr/80/index.adoc
```diff
@@ -1,22 +1,90 @@
...
+
+## Current Architecture
+
+None
+
+## Proposed Architecture
+
+https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used in this proposal.
+
+It is RECOMMENDED that managed services use https://source.redhat.com/groups/public/openshiftplatformsre/wiki/backplane_scripts[Back Plane]
```
tombentley, line 47: Non-public link

_adr/80/index.adoc
```diff
@@ -1,22 +1,90 @@
...
+* Identify the preferred solution for the diagnostic gathering mechanism.
+* Provide guidelines for content that should be included.
+
+## Non-goals
+
+None
+
+## Stakeholders
+* Eng
+* BU
```
tombentley, line 35: The outside reader likely won't know what BU means (see #40).

_adr/80/index.adoc
k-wall: Fixed.

_adr/80/index.adoc
k-wall: Fixed.

_adr/80/index.adoc
k-wall: Agreed - I'll reword once the personas are agreed.

_adr/80/index.adoc
k-wall: I agree - this is impossible to guarantee.  I am trying to encourage the implementer of the ADR to think about what information the script collects  and take reasonable steps can be taken to avoid PII.  Let me reword

k-wall:
> This proposed ADR is interesting as it references extensively some Red Hat specific mechanism. I've started a conversation to see how we want to address this.

I've raised #43 to address this.

_adr/80/index.adoc
tombentley: It's still a non-public link, did you push your changes?

k-wall:
@StevenTobin @emmanuelbernard @tombentley would you mind reviewing the changes?

_adr/80/index.adoc
```diff
@@ -1,22 +1,89 @@
...
+The exact way way the script will interrogate state will be service specific, however if the service provided on top of OpenShift, it is
+RECOMMENDED that the script invoke the `oc adm inspect` command to gather Kubernetes state in addition to any other service specific
+state gathering commands.
+
+It is RECOMMENDED that the output from the backplane script be a UNIX tar stream.  This will allow the artifact to provide the diagnostic
+information in a structured form that can easily be processed using common command-line tools.
+
+It is RECOMMENDED that the script takes reasonable steps to avoid the collection of PII or other sensitive information. This may be done
+either by redacting artifacts known to contain secrets/keys etc, or not gathering certain information in the first place.  For
+example, in a Java based system where PII may be present in the Java heap, the safe option is not to collect heap dumps.
```
tombentley, line 63: While I understand the motivation for this, do we really understand and accept the implications? I know we don't have a good mechanism for collecting heap dumps right now, but this basically means no heap dumps, even when OOME has occurred. That rules out the main problem finding technique for a whole class of problem. This rule also prevents us from taking log dumps, for example, which again could severely limit our ability to fix problems, and understand root causes. 

I would much rather we had the contractual wording to collect this kind of diagnostic data if we really had to (even if it's not collected by default), even if we're obliged to delete the information within a certain timeframe of taking it. For example, what if there was a mechanism to identify files in the tar.gz which contained PII and those files were then managed in a way that was GDPR compliant? I assume you've investigated that route and found it's not possible, or we just can't go there yet, so I don't see this as a blocker, just a source of continuing headaches.

_adr/80/index.adoc
```diff
@@ -1,22 +1,89 @@
...
+## Proposed Architecture
+
+https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used in this proposal.
+
+It is RECOMMENDED that managed services use https://source.redhat.com/groups/public/openshiftplatformsre/wiki/backplane_scripts[Back Plane]
+https://github.com/openshift/managed-scripts/[managed scripts] as the foundation of the data gathering mechanism. Backplane has an
+established https://source.redhat.com/groups/public/openshiftplatformsre/wiki/backplane_cluster_permissions[authorization model] that ensures
+that members of the {glossary-customersupport} have the permissions to run the scripts.
+
+Each Managed Service that implements this ADR MUST provide one or more backplane scripts that interrogate the state of a service instance.
```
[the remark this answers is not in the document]
k-wall: My aim was not to be prescriptive about the number of scripts that would be delivered to support each service.  I'll adjust my phrasing to eliminate references to 'the script'.

_adr/80/index.adoc
k-wall: I think I understand your points. How about I reword along the lines of?

```suggestion
By default, it is RECOMMENDED that the scripts takes reasonable steps to avoid the collection of PII or other sensitive information. This may be done either by redacting artifacts known to contain secrets/keys etc, or not gathering certain information in the first place. The scripts MAY offer alternative modes of operations that collect deeper information.  The RECOMMENDED mechanism to enable these modes is to make use of [managed script parameters](https://github.com/openshift/managed-scripts/blob/main/scripts/SREP/example/metadata.yaml#L30).  If the use of such a mode may result in a artifacts containing PII or other sensitive information, the parameter's description MUST document that fact.
```

In the case where the user elects to use the special modes, the responsibility would be on them to treat the artefact in a manner compliant with the Organisation's PII data processing standards.

_adr/80/index.adoc
```diff
@@ -1,22 +1,89 @@
...
+The exact way way the scripts will interrogate state will be service specific, however if the service provided on top of OpenShift, it is
+RECOMMENDED that the scripts invoke the `oc adm inspect` command to gather Kubernetes state in addition to any other service specific
+state gathering commands.
+
+It is RECOMMENDED that the output from the scripts be a UNIX tar stream.  This will allow the artifact to provide the diagnostic
+information in a structured form that can easily be processed using common command-line tools.
+
+It is RECOMMENDED that the scripts takes reasonable steps to avoid the collection of PII or other sensitive information. This may be done
+either by redacting artifacts known to contain secrets/keys etc, or not gathering certain information in the first place.  For
+example, in a Java based system where PII may be present in the Java heap, the safe option is not to collect heap dumps.
```
k-wall, line 63:
```suggestion
By default, it is RECOMMENDED that the scripts takes reasonable steps to avoid the collection of PII or other sensitive information. This may be done either by redacting artifacts known to contain secrets/keys etc, or not gathering certain information in the first place. The scripts MAY offer alternative modes of operations that collect deeper information.  The RECOMMENDED mechanism to enable these modes is to make use of [managed script parameters](https://github.com/openshift/managed-scripts/blob/main/scripts/SREP/example/metadata.yaml#L30).  If the use of such a mode may result in a artifacts containing PII or other sensitive information, the parameter's description MUST document that fact.
```

_adr/80/index.adoc
k-wall: @tombentley would the proposed reword address your comment?

tombentley: Speaking as an engineer, "yes" that would scratch my itch of being able to get the information I might need in order to understand problems manifesting in production. 

But thinking about the PoV of some GDPR/InfoSec persona, "no" because I would want backplane to provide mechanisms for knowing when SREs use these options and ensuring that the information gathered through them were managed in accordance with national laws and organisational policies.

In other words, I don't think there can be easy technical solutions to this: We need to have a conversation with business, legal, SRE and dev representation to figure out what might be possible if backplane were able to have first-class support for collecting and managing sensitive information, or the acceptance that there will be circumstances in which we won't be able to fully investigate production problems (and be prepared to tell the end customer that "there's a problem, we don't understand it and our company policy on being able to collect possibly-sensitive data about the service we're running for you prevents us from getting to the root cause").

There _might_ be hard technical solutions to this. For example if customers' usage of the cluster is such that the topics contain only encrypted data (and the investigating engineers cannot access the keys) then there can be no leakage of sensitive data from doing a heap dump.

tombentley:
@k-wall what happening with this one?

_adr/80/index.adoc
grdryn: Not fixed, unresolving.

_adr/80/index.adoc
grdryn: Not fixed, unresolving

PR #20, opened 2022-02-08
feat: AP-9 Orchestration of services and resources dependencies

pmuir:


pmuir:
@emmanuelbernard here is the draft content from your doc

PR #27, opened 2022-02-08
feat: AP-2 Blast radius reduction principle

pmuir:


pmuir:
This writing up as a pattern, not a principle.

PR #28, opened 2022-02-08
feat: AP-12 How to push rules from git to production

pmuir:


pmuir:
I think we can make this into a pattern rather than a point decision by centering it around our use of GitOps.

PR #29, opened 2022-02-08
feat: add AP-11 Diagnosing a problem in production at the pod level

pmuir:


pmuir:
@emmanuelbernard I'm not convinced this is an architectural pattern. I think it's a howto - in which case it could be a blog or similar.

emmanuelbernard:
@StevenTobin can you see how we can convert this one into a SOP even though not actionable from a defined root cause yet? Would you agree.
We can also see how we would also blog it for exchange, @maxandersen fancy doing a blog around it?

emmanuelbernard:
@StevenTobin to see if the generic SPO can be improved with that content.

emmanuelbernard:
Closing as not able to find the content a productive home.

PR #33, opened 2022-02-08
Feat: add architectural patterns

pmuir:


_includes/header.html
```diff
@@ -23,6 +23,7 @@
         <a class="page-link" href="/context/">Context</a>
         <a class="page-link" href="/padr/">PADRs</a>
         <a class="page-link" href="/adr/">ADRs</a>
+        <a class="page-link" href="/ap/">APs</a>
```
tombentley, line 26: APs before ADRs, since there's a kind of hierarchy of applicability?

Also this forces the nav down below the site title in order to fit. We can probably make the About and Overview into a drop down menu, or make space (in another PR)?

tombentley (approved):
Minor layout nits which we can fix later. LGTM.

PR #39, opened 2022-02-09
Add ADR template

tombentley:


pages/about/adr-process.adoc
```diff
@@ -22,13 +22,20 @@ If so, it should also seek to identify.
...
+
+. Fork {{site.github.repository_url}}[this repo] on github
+. Check out your fork (`git clone _YORKFORK_ && cd architecture`)
+. Create a branch (`git checkout -b adr-__nnnn__`)
+. Create your ADR file from the template (`cp -r _adr/0 _adr/__nnnn__`)
+. Update your ADR file:
+  * If the issue identified any applicable PADRs the `applies_padrs` property should be added to the front matter. 
   It's an array of all the applicable PADRs.
-* If the issue identified and tags, the `tags` property should be added to the front matter. 
+  * If the issue identified and tags, the `tags` property should be added to the front matter.
```
k-wall, line 35: Not really sure what this is trying to say.

tombentley: This better?

```suggestion
  * Appropriate `tags` should be added to the front matter. 
```

It's a little vague because while people seem to like the tags idea, we're not quite sure what tags we should have, or quite how they will be used. I can imagine in a few years when there are a few hundred ADRs tags will be useful for filtering the index in order to find the decisions that affected given components, for example. Let me flip it around and ask what tags you think we should have?

pmuir: Yes, better

pmuir (approved):

PR #42, opened 2022-02-15
ADR-81: Kafka instance scaling

tombentley:


bibryam (approved):
I'm guessing storage is not part the contract discussed here and that is why not mentioned. 
For completeness, there is also connection attempts property.

_adr/81/index.adoc
```diff
@@ -26,3 +26,152 @@ Excluding storage sizing, the Kafka service currently supports a single size of
...
+### Fleet manager/fleet shard interface
+
+Principles:
+
+* A fleet shard is able to provide a subset of series, possibly only one. 
+* The availability of a series might be limited to a particular cloud, region or combination of both.
+* It is a private concern of the shard exactly how each size in the series is provisioned.
+    * In principle it could use vertical, horizontal scaling and density > 1, in addition to applying quotas to limit capacity to the advertised limits.
+* When the manager provides an ConfigMap of the functional parameters the shard is expected to provide.
+* Within the shard image it has a model for how to provision each set of functional parameters.
```
rareddy, line 130: I did not follow these bullet points

emmanuelbernard (changes requested):

_adr/81/index.adoc
```diff
@@ -26,3 +26,152 @@ Excluding storage sizing, the Kafka service currently supports a single size of
...
+The map keys will be the series, and the elements of the list will be the available sizes within that series, in increasing order. 
+Nothing is implied by the order of the keys in the series map, e.g. an "ex2" may occur before or after an "ex1" and the order doesn't imply anything about those series.
+
+Reusing the example series defined above:
+[source,json]
+----
+"sizes": {
+  "ex1": [
+    { 
+      "name": "small",
```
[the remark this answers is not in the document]
gunnarmorling: Same for regions and the other availability limitations you mention below?

gunnarmorling (commented):
Some minor comments/suggestions inline.

_adr/81/index.adoc
```diff
@@ -19,10 +19,243 @@ NOTE: https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used
...
+
+The supported series are passed to the fleet manager via app interface.
+
+When terraforming a shard, the manager needs to keep track of which series are/will be supported on that shard.
+
+Currently it is acceptable for the manager to embed knowledge of the machine types needed by the shard.
+The manager does not know how the nodes running on those machines will be used.
+It is expected that a future ADR will describe a mechanism for the shard and manager to dynamically adjust the number of nodes.
+
+=== Fleet manager/fleet shard interface
```
tombentley, line 194: @rareddy I've elaborated quite a bit on this whole section now. Can you take another look?

@shawkins please also take a look.

_adr/81/index.adoc
```diff
@@ -26,3 +26,152 @@ Excluding storage sizing, the Kafka service currently supports a single size of
...
+        "partitions": "1000",
+        "connections": "500",
+        "connectionRate": "100",
+        "maxMessageSize": "1MB"
+      }
+    },
+    { 
+      "name": "medium",
+      "parameters": {
+        "ingress": "60MB/s",
```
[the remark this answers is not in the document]
wtrocki: @gunnarmorling amazing point! I think if that would be possible it will be good to have it separated.
It will be more future proof as UI might use different shortcuts/full names of the units.

That will be also good pattern for validation purposes (although we only use partitions value for validation now)

Reference for UI:
https://github.com/bf2fc6cc711aee1a0c2a/kafka-ui/blob/main/src/constant/constants.ts#L19


https://github.com/bf2fc6cc711aee1a0c2a/kafka-ui/blob/main/src/constant/constants.ts#L19

tombentley:
@emmanuelbernard, @gunnarmorling I've elaborated it quite a bit if you want to take another look.

_adr/81/index.adoc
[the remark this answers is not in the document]
riccardo-forina: Yes please keeping them separate would also allow for easier internationalization of the UoM (not saying we'll have to do that, I have no idea if MB/s can be written differently in other languages, but all the above pros still apply)

tombentley:
@pb82 please can you take a look at this too?

_adr/81/index.adoc
```diff
@@ -19,10 +19,243 @@ NOTE: https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used
...
+
+=== Fleet manager responsibilities
+
+The supported series are passed to the fleet manager via app interface.
+
+When terraforming a shard, the manager needs to keep track of which series are/will be supported on that shard.
+
+Currently it is acceptable for the manager to embed knowledge of the machine types needed by the shard.
+The manager does not know how the nodes running on those machines will be used.
+It is expected that a future ADR will describe a mechanism for the shard and manager to dynamically adjust the number of nodes.
```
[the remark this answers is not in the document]
davidffrench: I think it is reasonable to keep it vague since we are saying it will be described in a future ADR. What do you feel is missing here that could add context to this ADR?

_adr/81/index.adoc
```diff
@@ -19,10 +19,243 @@ NOTE: https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used
...
+
+.Sequence diagram of UI/Fleet manager interaction
+image::ui-fleet_manager.png[UI/Fleet manager interaction]
+
+After the UI has discovered the available cloud providers and regions it would make another GET request `/api/kafkas_mgmt/v1/series/{cloudId}/{cloudRegion}` to discover the series available in this cloud and region.
+An example response JSON is shown below.
+
+[source,json]
+.Example series available to a UI from the Fleet Manager's `GET /api/kafkas_mgmt/v1/series/{cloudId}/{cloudRegion}` endpoint
+----
```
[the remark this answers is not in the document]
pmuir: I wasn't expecting the sizes to be per cloud provider/region, I thought they were consistent across all cloud providers/regions from what @bibryam said.

_adr/81/index.adoc
```diff
@@ -19,10 +19,243 @@ NOTE: https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used
...
+            { "name": "connectionRate", "value": "100", "unit": "/s" }
+            { "name": "maxMessageSize", "value": "1", "unit": "MB" }
+          ]
+        }
+      ],
+    },
+    // other series as necessary
+]
+----
+<1> This list is not considered ordered
```
pmuir, line 143: It needs to be ordered so that UI can show it in the desired order (as I don't think the natural order is correct).

_adr/81/index.adoc
```diff
@@ -19,10 +19,243 @@ NOTE: https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used
...
+<2> Names MUST be unique within the `series` list 
+<3> The `id` uniquely identifies a series, allowing the name to be changed
+<4> Can be used in a UI/CLI to help guide users to appropriate profile/flavour of the service
+<5> The list MUST be ordered such that later items represent clusters with more capacity
+<6> The name MUST be unique within the `sizes` list
```
pmuir, line 148: A description field might be useful here, What do you think? @jgiardino

_adr/81/index.adoc
```diff
@@ -19,10 +19,243 @@ NOTE: https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used
...
+----
+"series": [ <1>
+    { "name": "Example", <2>
+      "id": 1, <3>
+      "description": "Multi-broker, multi-AZ", <4>
+      "sizes": [ <5>
+        { 
+          "name": "small", <6>
+          "parameters": [ <7>
+            { "name": "ingress", "value": "30", "unit": "MB/s" }
```
pmuir, line 108: just a note, we are using binary prefixes for units (i.e. MiB)

_adr/81/index.adoc
```diff
@@ -19,10 +19,243 @@ NOTE: https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used
...
+An example response JSON is shown below.
+
+[source,json]
+.Example series available to a UI from the Fleet Manager's `GET /api/kafkas_mgmt/v1/series/{cloudId}/{cloudRegion}` endpoint
+----
+"series": [ <1>
+    { "name": "Example", <2>
+      "id": 1, <3>
+      "description": "Multi-broker, multi-AZ", <4>
+      "sizes": [ <5>
```
pmuir, line 104: probably want a `defaultSize` field for each series.

_adr/81/index.adoc
```diff
@@ -19,10 +19,243 @@ NOTE: https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used
...
+
+The service does not offer any scaling options, except for storage, which is orthogonal to this problem.
+
+== Proposal
+// What is the decision being proposed
+
+=== Conceptual model
+Conceptually we define a _series_, which is an ordered list of _sizes_ of instance. 
+
+A _series_ has a name, and a description. The description might include what is common across the sizes it supports (e.g. "3 AZ").
```
pmuir, line 54: I'm not sure "series" is the most intuitive name here - it took me minutes to realize what it was.

_adr/81/index.adoc
```diff
@@ -19,10 +19,243 @@ NOTE: https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used
...
+
+A _size_ has a name and a set of <parameter, value> pairs. Each of the sizes in a series MUST have the same set of function parameters (e.g. ingress bandwidth), but the values MAY differ.
```
pmuir, line 56: why does a series have a name and an id, but a size only a name?

pmuir (changes requested):
Mostly minor API shape issues - the reason I am requesting changes is around whether series are per cloud provider/region, or global.

_adr/81/index.adoc
tombentley: I'm not wedded to it, but any suggestions?

emmanuelbernard: T-shirts! /Me run away.

_adr/81/index.adoc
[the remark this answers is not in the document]
tombentley: There's nothing to stop us serving the same set of series from `.../aws/us-east-1` as from some other cloud and region. My thinking for parameterising the endpoint on provider and region was this would allow a future flavour of the service which ran on e.g. ARM silicon and local SSD from those AWS regions which supported it, and for the UI to make that apparent. 

The main downside seems to be the need for an extra request by the UI.

emmanuelbernard (approved):

_adr/81/index.adoc
rareddy: - Streaming Unit Profiles
- Capacity Profiles

_adr/81/index.adoc
jgiardino: Are you thinking that a description could be something we display in the UI, or used for something else? I like the idea of having descriptions for these that we could present in the UI, and having them come from the API.  But I also realize that providing descriptions that are good requires extra work.

_adr/81/index.adoc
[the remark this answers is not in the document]
jgiardino: _My question is not related to @pmuir's comment above_  

I'm not familiar with the response we get with what's described here:
> After the UI has discovered the available cloud providers and regions it would make another GET request `/api/kafkas_mgmt/v1/series/{cloudId}/{cloudRegion}` to discover the series available in this cloud and region.

Does it make sense to include proposed updates to what would be returned with the list of available cloud providers and regions? Are we expecting any updates with that API call to provide details about available capacity in a region? 

Without seeing both responses (list of cloud providers/regions and list of sizes), it isn't clear to me if we're getting the information we need to communicate to the user something like “this configuration is temporarily unavailable, try a different configuration” before they submit the request.

_adr/81/index.adoc
tombentley: I'd much prefer a single word, and preferably a short one. But "profile" instead of series is fine imo?

_adr/81/index.adoc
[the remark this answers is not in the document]
tombentley: I guess I wanted some more formal wording to the effect that "The Fleet Manager does the business stuff, the Shard does the provisioning implementation stuff". 

I'm happy if you and the fleet manager people are happy with the responsibilities, both called out and implied, in this ADR.

_adr/81/index.adoc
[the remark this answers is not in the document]
tombentley: Currently the cloud and region are path parameters of the endpoint. I think that's cleaner that having to introduce cloud and region concepts into the schema, but it requires an extra round trip. There doesn't yet seem to be a consensus on this yet though.

tombentley:
I've switched from the term "series" to "profile", which reads a little clearer now. I'm not really wanting to get into the detailed API design, but I've also elaborated on the schema for profile discovery based on some of the feedback from @pmuir and @jgiardino.

_adr/81/index.adoc
tombentley: The need for a default arises from wanting to avoid a `v2` API. If we were happy to have v2 we'd just make it mandatory in the creation request. It feels a bit rubbish to need a default in the discovery response, just for that purpose. I feel like we could leave it out-of-band to the API, at least for now, since all existing clients of the API are only expecting to be able to create a "standard" instance anyway.

_adr/81/index.adoc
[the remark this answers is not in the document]
tombentley: @jgiardino do we do that right now? I know we were thinking about that kind of problem with the dynamic OSD sizing work, but that's on hold at the moment (we'll come back to it soon enough once we have support for larger instances).

I guess we _could_ use this endpoint to express which profiles and sizes were ready for immediate deployment. One problem is just how dynamic that information could be. There will always be the case where it turns out we can't provision an instance when handling an instance creation request. From an API PoV I think it might be nicer to support a `validateOnly` parameter on the creation endpoint, which the UI could poll regularly when the instance creation modal was displayed.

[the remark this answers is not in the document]
jgiardino: We are supposed to be getting information about whether a cloud provider region is available. I remember seeing some discussions on how to update that to support size selection, but I don't remember where.

Maybe @pb82 has more details about the API that provides details about cloud provider regions and how we would use that API along with the API for available sizes to communicate to users what their options are.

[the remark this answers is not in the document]
tombentley: This is a good point. I'd argue that availability for creating instances should be part of this new endpoint, since for a given cloud+region we might have capacity to create a 1SU/Small instance but not to create a 5SU/3Xlarge instance.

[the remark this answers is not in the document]
emmanuelbernard: is availability a _public_ concept (anyting the UI consumes has to be), or a private concept?

_adr/81/index.adoc
```diff
@@ -19,10 +19,264 @@ NOTE: https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used
...
+.Sequence diagram of UI/Fleet manager interaction
+image::ui-fleet_manager.png[UI/Fleet manager interaction]
+
+After the UI has discovered the available cloud providers and regions it would make another GET request `/api/kafkas_mgmt/v1/profile/{cloudId}/{cloudRegion}` to discover the profiles available in this cloud and region.
+An example response JSON is shown below.
+
+[source,json]
+.Example profile available to a UI from the Fleet Manager's `GET /api/kafkas_mgmt/v1/profiles/{cloudId}/{cloudRegion}` endpoint
+----
+"profiles": [ <1>
```
pb82, line 107: @pmuir do you think this format offers advantages for the UI (metadata) or do you think the current format described in https://docs.google.com/document/d/1VeZae6svCNVJS4XbeXL1q6EKdUh7Oe9qmJ4CDGV6rCA/edit#heading=h.m1mukm7lqaex are sufficient?

pmuir: The UI can handle either I think. For me it's more about which API we think is more understandable and better represents what we are doing. I'd argue this one does (except perhaps the name and desription fields are not useful as we want to localize them).

pb82: Is the top level `parameters` field useful? It defines names for all those parameters but I assume you will localization them too on the UI?

pb82: The only difference then is that sizes are a subarray. We could represent them in app-interface like this:

```yaml
    SUPPORTED_KAFKA_SIZES:
    - id: eval
      sizes: 
      -  id: x1
         ingressThroughputPerSec: “30Mi”
         egressThroughputPerSec: “30Mi”
         totalMaxConnections: 3000
         maxDataRetentionSize: “1000Gi”
         maxPartitions: 1000
         maxDataRetentionPeriod: “P14D”
         maxConnectionAttemptsPerSec: 100
    - id: standard
      sizes:
      -  id: x1
         ingressThroughputPerSec: “30Mi”
         egressThroughputPerSec: “30Mi”
         totalMaxConnections: 3000
         maxDataRetentionSize: “1000Gi”
         maxPartitions: 1000
         maxDataRetentionPeriod: “P14D”
         maxConnectionAttemptsPerSec: 100
```

pmuir: Yes, the UI can localize these based on the id. My comment was just whether the name and description fields were therefore useful (as they won't be used).

_adr/81/index.adoc
```diff
@@ -19,10 +19,264 @@ NOTE: https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used
...
+    { "id": 1, <2>
+      "name": "Example", <3>
+      "description": "Multi-broker, multi-AZ", <4>
+      "parameters": { <5>
```
pb82, line 111: @pmuir are these useful to the UI?

_adr/81/index.adoc
tombentley: Ah, OK. Out of interest, is there consistent l10n in the CLI @wtrocki ?

wtrocki: Yes.

tombentley: I've tweaked the ADR, removing the `name` and `description` fields. I've also added "cost" as a parameter @pb82 pointed out that that was needed at least by the fleet manager so it knows about the AMQ quota cost. I'm thinking this is something that the UI also perhaps needs to surface, and therefore it should be in this API.

_adr/81/index.adoc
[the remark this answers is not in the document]
tombentley: Just to be clear, there two things going on here. 

First there is the capability problem: Whether a given cloud + region can support a given size in principal. I.e. we've decided to enable that size for that cloud and region, which is a business decision. We want the ability to limit at least _profiles_ to a cloud and region. e.g. when we roll out support for some enterprisey profile we probably don't want to do that on Azure and AWS _simultaneously_. Or if we ever had a profile that was tied to AWS infra that wasn't available in all regions that we operated then we'd need to be able to express that too.

Then there is the capacity problem: whether we can provision an instance of a certain size _right now_.

I believe both must (eventually) be a public concepts. The UI must be able to inform the user what is possible and what is possible right now. 

The contract being outlined here is about expressing the capability part. We will eventually need to deal with the capacity part. That could be do using the `validateOnly` parameter on the creation endpoint I mentioned, but TBH I see it as out of scope for this ADR.

tombentley:
I've made a few final tweaks to this ADR today, based on the conversation. I've used the 'real' profile and size names, because it seems people mostly found the more abstract terms confusing. The comments so far have been really useful (thanks!). I've marked the PR as ready for review, so if you have any further questions or comments now is the time to voice them.

Assuming no gotchas, I'd like to merge this PR by the end of this week.

_adr/81/index.adoc
```diff
@@ -19,10 +19,267 @@ NOTE: https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used
...
+        "maxMessageSize": { "name": "Maximum message size", "unit": "MiB" },
+        "cost": { "name": "Number of entitlements required to create an instance" }
+      },
+      "sizes": [ <4>
+        { "id": "s", <5>
+          "parameterValues": { <7>
+            "ingressThroughputPerSec": "30", "egressThroughputPerSec": "30", 
+            "maxPartitions": "1000", "totalMaxConnections": "500",
+            "maxConnectionAttemptsPerSec": "100", "maxMessageSize": "1",
+            "cost": 1,
```
pb82, line 124: The only concern about this is that `cost` is an internal detail of the fleet manager (Quota cost). I'm not sure if we should expose it here. What do you think @ziccardi ?

tombentley: Ah, well if it's internal then no need to expose it in the contract between the UI and the fleet manager. That doesn't prevent us passing it into the fleet manager from app interface though. But I guess I'd still like to know whether/how a notion of cost gets into the UI. @pmuir probably knows.

ziccardi: We have to put the size somewhere: we can enforce some naming on the sizes id ('x1', 'x2', etc) or add a `cost` attribute: IMO we will expose the detail either way.
Having to choose one, I'd prefer to have the `cost` attribute.

tombentley: I think that's fine for the app-interface/fleet manager contract, and omit that parameter in the contract between fleet manager and its clients.

_adr/81/index.adoc
```diff
@@ -19,10 +19,267 @@ NOTE: https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used
...
+`medium`:: Ingress 60MB/s, egress 60MB/s, 2000 partitions, 1000 connections per instance, max 100 connections per second, 1MB max message size.
+`large`:: Ingress 90MB/s, egress 90MB/s, 3000 partitions, 1500 connections per instance, max 100 connections per second, 1MB max message size.
+
+In this example, the functional attributes scale linearly, with the exception of connection rate and max message size.
+In general, linearity SHALL NOT be a requirement, even for those attributes which do scale up.
+For example, it would have been acceptable to define the `large` ingress as 100M/s, even though this is not 3× the `small` ingress, even when the other `large` functional parameters are 3× their `small` values.
+
+=== General principles
+
+* An individual fleet shard is required to provide only a subset of all profiles supported by the service globally.
```
pb82, line 74: That would mean the Fleet Manager needs to factor in the supported profiles when picking a cluster for placement. How can the fleet manager know which profiles are supported by the fleet shard? Or is this purely a matter of fleet isolation? In that case, do we really need profile when we already have instance type?

_adr/81/index.adoc
```diff
@@ -19,10 +19,267 @@ NOTE: https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used
...
+The `POST /api/kafkas_mgmt/v1/kafkas` endpoint would change to take the profile id and size id in addition to the existing parameters.
+
+[source,json]
+.Example `POST /api/kafkas_mgmt/v1/kafkas` made by a UI to the Fleet Manager to create an instance
+----
+{
+    "region": "us-east-1",
+    "cloud_provider": "aws",
+    "profileId": "standard", <1>
+    "sizeId": "s", <2>
```
pawelpaszki, line 189: @tombentley - just to confirm, based on the discussion held earlier today - UI will send one parameter (called `plan`), in format `<instance_type>.<size_id>`. is that correct?

tombentley: I think @pmuir was happy with either (as am I).

_adr/81/index.adoc
```diff
@@ -19,10 +19,285 @@ NOTE: https://datatracker.ietf.org/doc/html/rfc2119[RFC-2119] keywords are used
...
+      "parameters": { <3>
+        "ingressThroughputPerSec" : { "name": "Ingress", "unit": "MiB/s" },
+        "egressThroughputPerSec": { "name": "Egress", "unit": "MiB/s" },
+        "maxPartitions": { "name": "Partitions" },
+        "totalMaxConnections": { "name": "Connections" },
+        "maxConnectionAttemptsPerSec": { "name": "Connection Rate", "unit": "/s" },
+        "maxMessageSize": { "name": "Maximum message size", "unit": "MiB" },
+        "quotaType": { "name": "The kind of quota an instance of this plan consumes" }, <4>
+        "quotaConsumed": { "name": "The amount of quota instance of this plan consumes" }, <5>
+        "capacityConsumed": { "name": "An amount of shard capacity consumed by an instance of this plan" } <6>
```
tombentley, line 121: @pmuir @pb82 I think this is what we agreed, right?

I don't _particularly_ like how these three are magically common to all instance types, whereas there's no requirement for other parameters to be, but I'm not sure defining something parallel to parameters is really worth while.

tombentley:
@pb82 @JameelB @davidffrench @pmuir @ziccardi @pawelpaszki I've pushed a couple more commits based on the meeting we had earlier today. Please check that the text matches your understanding from that meeting.

_adr/81/index.adoc
pb82: Yes, looks good to me. And what we will use in app-iterface: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/34777/diffs#8b58cbce883c482be4c1ce504427b2d7776e236c_277_256

PR #46, opened 2022-03-22
Import ADR-76

grdryn:
Title: Handling usage of dynamically-async-created DNS records from
dependent services

### DIFF
--- a/_adr/76/index.adoc
+++ b/_adr/76/index.adoc
@@ -1,0 +1,96 @@
+---
+num: 76
+title: "Handling usage of dynamically-async-created DNS records from dependent services"
+status: "Accepted"
+authors:
+  - "Gerard Ryan"
+tags:
+  - kafka
+applies_padrs: # What PADRs does this ADR apply?
+applies_patterns: # What APs does this ADR apply?
+---
+
+:link-adr-57: link:../57/index.html[ADR-57]
+:abbr-rhosak: pass:[<abbr title="Red Hat OpenShift Streams for Apache Kafka">RHOSAK</abbr>]
+// Top style tips:
+// * Use one sentence per line
+// * No unexpanded acronyms
+// * No undefined jargon
+
+// No need for a title heading, it's added by the template
+
+== Context and Problem Statement
+// What is the background against which this decision is being taken?
+
+For some services (such as {abbr-rhosak}), when a new instance is being created for a customer, we create new cluster-agnostic DNS records in Route53 via the control plane, pointing to resources that the customer needs access to.
+We also use those customer-facing domains to monitor availability (such as via the Strimzi Canary in the {abbr-rhosak} case).
+
+The problem is that the data plane components process may start continuously trying to connect the domain for a period of time before the domain has been created.
+This can lead to an issue called “negative DNS caching”, where DNS resolvers, between the querying process and the authoritative DNS servers for the domain, cache the non-existence of the domain for a few minutes.
+This can lead to a longer lead time to the monitoring process reporting availability metrics.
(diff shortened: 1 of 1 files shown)


_adr/76/index.adoc
```diff
@@ -0,0 +1,96 @@
...
+authors:
+  - "Gerard Ryan"
+tags:
+  - kafka
+applies_padrs: # What PADRs does this ADR apply?
+applies_patterns: # What APs does this ADR apply?
+---
+
+:link-adr-57: link:../57/index.html[ADR-57]
+:abbr-rhosak: pass:[<abbr title="Red Hat OpenShift Streams for Apache Kafka">RHOSAK</abbr>]
```
grdryn, line 14: I haven't been able to build the site on Fedora (it doesn't have a Ruby version old enough for this version of Jekyll) to see how well these attributes work when rendered. Managed to build on [ubi8/ruby-27](https://catalog.redhat.com/software/containers/ubi8/ruby-27/5f521b90acb94fdd904e98d7), but couldn't run (I gave up after troubleshooting past the first 2 or 3 issues).

_adr/76/index.adoc
```diff
@@ -0,0 +1,96 @@
...
+// Top style tips:
+// * Use one sentence per line
+// * No unexpanded acronyms
+// * No undefined jargon
+
+// No need for a title heading, it's added by the template
+
+== Context and Problem Statement
```
grdryn, line 22: Deviated from [template](https://github.com/bf2fc6cc711aee1a0c2a/architecture/blob/main/_adr/0/index.adoc) slightly here for headings, to use the more Asciidoc notation of `==` instead of `##` (I think the latter works because of some Asciidoctor compatibility for that Markdown syntax—a different processor might not). I can create a PR to update the template if that would be useful too.

_adr/76/index.adoc
```diff
@@ -0,0 +1,96 @@
+---
+num: 76
+title: "Handling usage of dynamically-async-created DNS records from dependent services"
+status: "Accepted"
```
grdryn, line 4: This was already accepted here: https://docs.google.com/document/d/1ypysP8u4eP7zvsYVScYi2SlRNwixXHrbazyiOoorQqA/edit

Is there an expected way to link the old location, for preservation of history (at least for those who have access to the old location)?

grdryn (commented):
@tombentley I've added a few questions here, I think you might be the best person to give authoritative answers? Thanks!

_adr/76/index.adoc
grdryn: Created issue #47 for the issues I was having with building on Fedora.

grdryn:
@tombentley @emmanuelbernard please take a look, I think this can be merged. Thanks! It's referred to in #45 as an example of AP-13.

_adr/76/index.adoc
tombentley: A PR to update the template would be great. Thanks!

_adr/76/index.adoc
tombentley: No I don't think we need to preserve the old location. People can always look it up by the number.

tombentley (approved):
LGTM, thanks. I didn't check it was the same words as the accepted gdoc version, but I trust you :smile:

grdryn:
Thanks for the approvals folks. I don't have the ability to merge here, so one of you will have to do it.

I still have some follow-up TODO items for some of the stuff I mentioned above. I will get around to those soon, but those will be separate PRs, and none of them should block this from being merged, in my opinion.

PR #51, opened 2022-03-28
add adr-82: streaming unit hours

pb82:
ADR-81: Streaming Unit Hours

This ADR proposed a way of calculating the cost per streaming unit hour in a way that works for different Kafka sizes. The proposal is based on metrics.

### DIFF
--- a/_adr/82/index.adoc
+++ b/_adr/82/index.adoc
@@ -1,0 +1,87 @@
+---
+num: 82
+title: Streaming Unit Hour Metrics
+status: "Draft"
+authors:
+  - "Peter Braun"
+tags:
+  - "kafka"
+#applies_padrs: # What PADRs does this ADR apply?
+#applies_patterns: # What APs does this ADR apply?
+---
+
+## Context and problem statement
+
+Streaming Unit pricing will be accomplished by introducing a new concept called ‘Streaming Unit Hours’. This is defined as Cluster Hours multiplied by the Streaming Unit Size. Here Streaming Unit Size should not be confused with Quota usage which is required for prepaid pricing.
+
+Cluster Hours are currently recorded on the data plane clusters using a recording rule (kafka_id:strimzi_resource_state:max_over_time1h). We need a way to multiply the value of this with the Streaming Unit Size for a given instance.
+
+Other metrics don’t need to be multiplied: storage and traffic are billed independently of the Streaming Unit.
+
+## Goals
+
+* Modify the recording rule kafka_id:strimzi_resource_state:max_over_time1h to multiply its value by the Streaming Unit Size.
+
+## Non-goals
+
+* Changes on the Subwatch side. This ADR is only concerned with the Managed Kafka changes.
+* Define additional fields to hold the cost per Streaming Unit hour. The amount is simply multiplied by the instance size.
+
+Stakeholders
+
+* Eng
+* BU
+* QE
+
+## Current architecture
+
+Cluster Hours are defined in our link:https://github.com/bf2fc6cc711aee1a0c2a/observability-resources-mk[Observability Resources] repository among other consumption related metrics. It is collected on the data plane and remote written to Observatorium. From there it is made available to Subscription Watch who use it to create invoices.
+
+Cluster Hours is based on the maximum-per-one-hour value of the underlying strimzi_resource_state metric which is 1 when a Kafka CR is present in the namespace. Thus, our billing is accurate to the hour.
+
+The Kafka instance size in Streaming Units is not taken into account at any point here.
+
+## Proposed architecture
+
+Proposed Architecture
+
+*Firstly*, the size in Streaming Units needs to be available to the data plane. That is currently not the case. The options to achieve this are:
+
+1. Include it in the Managed Kafka CR as an annotation. The CR is retrieved by the Fleetshard sync component via the /api/kafkas_mgmt/v1/agent-clusters/{id}/kafkas endpoint.
+2. Include it in the Managed Kafka CR Capacity spec. Also retrieved by the Fleetshard sync component via the /api/kafkas_mgmt/v1/agent-clusters/{id}/kafkas endpoint.
+
+*Secondly*, the Streaming Unit size needs to be exposed as a metric on the data plane. Components that could expose this are either the Fleetshard Operator, the Fleetshard Sync component or the Canary.
+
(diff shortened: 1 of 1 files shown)


_adr/82/index.adoc
```diff
@@ -0,0 +1,87 @@
...
+The Kafka instance size in Streaming Units is not taken into account at any point here.
+
+## Proposed architecture
+
+Proposed Architecture
+
+*Firstly*, the size in Streaming Units needs to be available to the data plane. That is currently not the case. The options to achieve this are:
+
+1. Include it in the Managed Kafka CR as an annotation. The CR is retrieved by the Fleetshard sync component via the /api/kafkas_mgmt/v1/agent-clusters/{id}/kafkas endpoint.
+2. Include it in the Managed Kafka CR Capacity spec. Also retrieved by the Fleetshard sync component via the /api/kafkas_mgmt/v1/agent-clusters/{id}/kafkas endpoint.
```
machi1990, line 51: Won't including it as an annotation suggested above suffice?

_adr/82/index.adoc
```diff
@@ -0,0 +1,87 @@
...
+
+*Secondly*, the Streaming Unit size needs to be exposed as a metric on the data plane. Components that could expose this are either the Fleetshard Operator, the Fleetshard Sync component or the Canary.
+
+Why a metric and not a label? It does not seem to be possible to multiply a Prometheus metric with a value retrieved from a label. Instead, the multiplication needs to happen as a join of two metrics, the existing Cluster Hours and the new Streaming Unit Size.
```
tombentley, line 55: "It does not seem to be possible" it would be good to call out exactly the component that doesn't support this.

_adr/82/index.adoc
```diff
@@ -0,0 +1,87 @@
...
+
+*Thirdly*, the new Streaming Unit Hours metric needs to be scraped and made available to the cluster monitoring stack. It does not need to be sent to Observatorium as it will only be used in the recording rule to calculate the Cluster Hours (now Streaming Unit Hours). The query in the recording rule needs to be updated to multiply the value by the Streaming Unit size,e.g.
```
tombentley, line 57: Can we clarify where the recording rule lives? It's in the dataplane OSD, right?

_adr/82/index.adoc
```diff
@@ -0,0 +1,87 @@
...
+  - "Peter Braun"
+tags:
+  - "kafka"
+#applies_padrs: # What PADRs does this ADR apply?
+#applies_patterns: # What APs does this ADR apply?
+---
+
+## Context and problem statement
+
+Streaming Unit pricing will be accomplished by introducing a new concept called ‘Streaming Unit Hours’. This is defined as Cluster Hours multiplied by the Streaming Unit Size. Here Streaming Unit Size should not be confused with Quota usage which is required for prepaid pricing.
```
tombentley, line 15: Conceptually there's a difference between passing a number that is _defined_ to be the number of streaming units, and passing a factor/coefficient that _happens to be_ the number of streaming units. The former makes linearity of cost something that's built into the architecture. The latter does not. While we think that's fine right now (i.e kind of a built-in assumption of streaming units) we may find that as we scale up we want to discount larger clusters. For example, a 3 broker cluster also needs three zk/kraft controller nodes, but a 12 broker cluster still only needs three zk/kraft controller nodes, so the VM _costs_ do not scale linearly, even if what we charge may do.

_adr/82/index.adoc
```diff
@@ -0,0 +1,87 @@
...
+
+Other metrics don’t need to be multiplied: storage and traffic are billed independently of the Streaming Unit.
+
+## Goals
+
+* Modify the recording rule kafka_id:strimzi_resource_state:max_over_time1h to multiply its value by the Streaming Unit Size.
+
+## Non-goals
+
+* Changes on the Subwatch side. This ADR is only concerned with the Managed Kafka changes.
```
tombentley, line 27: meta: Should the ADR talk in terms of subwatch, or should a less RH-specific term be used?

_adr/82/index.adoc
```diff
@@ -0,0 +1,87 @@
...
+
+Stakeholders
+
+* Eng
+* BU
+* QE
+
+## Current architecture
+
+Cluster Hours are defined in our link:https://github.com/bf2fc6cc711aee1a0c2a/observability-resources-mk[Observability Resources] repository among other consumption related metrics. It is collected on the data plane and remote written to Observatorium. From there it is made available to Subscription Watch who use it to create invoices.
```
tombentley, line 38: meta: Observatorium, or generic placeholder?

_adr/82/index.adoc
```diff
@@ -0,0 +1,87 @@
...
+
+* No threat model changes are expected here.
+
+## Requires architecutral overview update
+
+* No architectural overview update anticipated.
+
+## Alternatives considered / rejected
+
+Alternatives to the described approach are multiple products and Streaming Unit metadata. They have been elaborated in link:https://docs.google.com/document/d/1fkeEqZ6JIclsgtynssBH43uphP2dVgvuPmsb-eaDKro/edit#heading=h.gn81l1rechyx[this document].
```
tombentley, line 75: Is there a reason we can't include that here?

emmanuelbernard:
Hey @pb82 Tom and Manyanda provided some feedback. When you're back, can you take a look at these to move forward?

_adr/82/index.adoc
pb82: Yes, that should be fine. I just wanted to list all options but we can say that the annotation is preferred.

machi1990: Thanks.

_adr/82/index.adoc
pb82: The discount could come from the control plane, by setting the number to e.g. `2.5` instead of `3`. You're right that his shouldn't be called streaming hours then. How about `billed_streaming_units`?

pb82:
@tombentley @machi1990 I've addressed your comments

machi1990 (approved):

_adr/82/index.adoc
tombentley: Part of the point of using a github repo and a website is to make this stuff accessible to people outside Red Hat. So I would prefer to include that content inline here unless there's a good reason not to, and I'm not sure efficiency is (but I can't actually see that content, so maybe it's really hard to replicate in asciidoc, but then my question would be: Can be make it readable outside the company).

_adr/82/index.adoc
```diff
@@ -0,0 +1,89 @@
+---
+num: 82
+title: Streaming Unit Hour Metrics
+status: "Draft"
```
tombentley, line 4: @pb82 can you change this to "Accepted" before I merge it?

tombentley (approved):

PR #52, opened 2022-03-31
add adr-83 permission api

### LINKED ISSUE #48: [New ADR]: KAS-FLEET-MANAGER Permissions Check API
### What
We need to write a new ADR about a _Permissions Check API_ for the KAS-FLEET-MANAGER.
Such API will be similar to the `oc canI` api and will allow the caller to query the fleet manager about what he can or can't do and why.

### Why
Currently the only way to know if an action is allowed or not is by trying and checking the outcome.
Some of our APIs consumer (for example, the UI) needs to know upfront making the call if the call will be successful or not (to enable buttons, populate comboboxes, sliders, etc.): that means that without a permission API, they have to replicate **exactly** the same checks we already do in KAS-FLEET-MANAGER and every time something changes into the KAS-FLEET-MANAGER, they have to update their code accordingly.

### How
The _Permissions Check API_ should be something like this:

```
    /can/{verb}/{resource}
```
where `verb` will be the action and `resource` will be the _resource_ affected by the _action_.

The output of the call will be a json containing an `outcome` (granted or denied) and some details:

- if the outcome is `granted`, the detail will describe what has been granted
- it the outcome is `denied`, the detail will contain a list of errors describing what the user miss to be able to perform the action

#### Create Kafka

```
GET /can/create/kafka
```

**Granted Outcome:**
```json
{
   "outcome":"granted",
   "details":{
      "plans":[
         "standard.x1",
         "standard.x2",
         "eval.x1"
      ],
      "quota":{
         "granted":6,
         "available":4
      }
   }
}
```

**Denied Outcome - terms not accepted**
```json
{
   "outcome":"denied",
   "errors":[
      {
         "code":"ERR-001",
         "description":"Required terms have not been accepted",
         "metadata":[
            {
               "termsurl":"https://url.to.accept.terms"
            }
         ]
      }
   ]
}
```

ziccardi:
ADR-83: Permission API

This ADR propose a way for the Kas Fleet Manager API consumers to check in advance whether a certain request would be successful.

Closes #48

### DIFF
--- a/_adr/83/index.adoc
+++ b/_adr/83/index.adoc
@@ -34,24 +34,24 @@ currently UI and CLI have to perform at least the following checks:
 Another point of attention is that currently, any change in the flow requires an update to all the actors/consumers who are carrying out those checks.
 
 ## Goals
-* API consumers can query a new KAS Fleet Manager Permission API to check if they can perform a certain action and receive a response with the outcome (allowed/denied) containing all the granted resources or, eventually, the missing requirements.
+* API consumers can query a new KAS Fleet Manager Can-I API to check if they can perform a certain action and receive a response with the outcome (allowed/denied) containing all the granted resources or, eventually, the missing requirements.
 
 ## Non-goals
 
 ## Current architecture
-* Kas Fleet Manager currently only allows the ‘try and see if it works’ approach: if what you are trying to do fails, Kas Fleet Manager gives you some details on went wrong, but you have no way of checking in advance if what you want to do is possible
-* API Consumers (like UI) are doing the required checkss into their codebase and every time a change occurs, both Kas Fleet Manager and UI have to be updated to reflect the change
+* Kas Fleet Manager currently only allows the ‘try and see if it works’ approach: if what you are trying to do fails, Kas Fleet Manager gives you some details on what went wrong, but you have no way of checking in advance if what you want to do is possible
+* API Consumers (like UI) are doing the required checks into their codebase and every time a change occurs, both Kas Fleet Manager and UI have to be updated to reflect the change
 
 ## Proposed architecture
 * We need to create a new API.
-The API will be authenticated and will allow the caller to check in advance whether he can perform a certain operation.
-A different endpoint will be provided for each verb-resource pair, wich, in turn, will provide a schema defining exactly what the answer will look like.
+The API will be authenticated and will allow the callers to check in advance whether they can perform a certain operation.
+A different endpoint will be provided for each verb-resource pair, which, in turn, will provide a schema defining exactly what the answer will look like.
 +
 ```
 GET /api/v1/can/{verb}/{resource}?deep=true
 ```
 +
-The API will stop at the first error unless `deep=true` is specified
+The API will stop at the first error unless `deep=true` is specified.
 Output of the API will be a JSON with at least the following content:
 +
 ```
@@ -112,7 +112,7 @@ Some real world example for an invocation of the `/can/create/kafka` could be:
 }
 ```
 
-.. Terms not accepted and insufficient quota
+.. Terms not accepted and insufficient quota (with deep=true)
 +
 ```
 {
@@ -146,7 +146,7 @@ Some real world example for an invocation of the `/can/create/kafka` could be:
 * Client application will have to be updated to take advantage of the new API
 
 ## Dependencies
-* Current check logic implemented in the Kas Fleet Manager needs to be moved to a common, reusable framework that will be used by both the Kas Fleet Manager middlewares and the new Permission API
+* Current check logic implemented in the Kas Fleet Manager needs to be moved to a common, reusable framework that will be used by both the Kas Fleet Manager middlewares and the new Can-I API
 
 ## Consequences if not completed
 


ziccardi:
@pb82 @wtrocki @tombentley @emmanuelbernard Mind giving a look please?

_adr/83/index.adoc
```diff
@@ -0,0 +1,131 @@
...
+title: Permission API
+status: "Draft"
+authors:
+  - "Massimiliano Ziccardi"
+tags:
+  - "kafka"
+#applies_padrs: # What PADRs does this ADR apply?
+#applies_patterns: # What APs does this ADR apply?
+---
+## Context and Problem Statement
```
emmanuelbernard, line 12: In a context of an eventual move to a KCP hosted API, how is that problem resolved? Is there such API? CC @pmuir

_adr/83/index.adoc
```diff
@@ -0,0 +1,131 @@
...
+* API consumers can query KAS Fleet Manager Permission API to check if they can perform a certain action and receive a response with the outcome (granted/denied) containing all the granted resources or, eventually, any missing requirements.
+
+## Non-goals
+
+## Current architecture
+* Kas Fleet Manager currently only allows the ‘try and see if it works’ approach: if what you are trying to do fails, Kas Fleet Manager gives you some details on went wrong, but you have no way of checking in advance if what you want to do is possible
+* API Consumers (like UI) are doing the required checkss into their codebase and every time a change occurs, both Kas Fleet Manager and UI have to be updated to reflect the change
+
+## Proposed architecture
+* We need to create a new API:
```
emmanuelbernard, line 42: I read quickly but can we clarified that this API will still be behind authentication so that it cannot be exploited anonymously or to check rights of others identities.

_adr/83/index.adoc
```diff
@@ -0,0 +1,131 @@
...
+    metadata: {
+      acceptance_url: "https://url.to.accept.terms"
+    }
+  }]
+}
+```
+
+### Threat model
+
+* No threat model changes are expected here.
```
emmanuelbernard, line 115: See my message on authentication. Such an API might increase the surface and thus threat.

emmanuelbernard (changes requested):

_adr/83/index.adoc
ziccardi: API will be authenticated and the user will be able to query only about his own permissions. Adding this to the ADR.

_adr/83/index.adoc
ziccardi: hmmm... The API won't expose anything new: it just aggregates in a single API info's that can already be gathered by querying different subsystems/api.

_adr/83/index.adoc
pmuir: Yes, kubernetes provides a SelfSubjectAccessReview API which is wrapped by `kubectl auth can-i` - which KCP supports - see https://kubernetes.io/docs/reference/access-authn-authz/authorization/#checking-api-access

pmuir: https://docs.openshift.com/container-platform/4.6/rest_api/authorization_apis/selfsubjectaccessreview-authorization-k8s-io-v1.html

_adr/83/index.adoc
```diff
@@ -0,0 +1,132 @@
...
+## Current architecture
+* Kas Fleet Manager currently only allows the ‘try and see if it works’ approach: if what you are trying to do fails, Kas Fleet Manager gives you some details on went wrong, but you have no way of checking in advance if what you want to do is possible
+* API Consumers (like UI) are doing the required checkss into their codebase and every time a change occurs, both Kas Fleet Manager and UI have to be updated to reflect the change
+
+## Proposed architecture
+* We need to create a new API.
+The API will be authenticated and will allow the caller to check in advance whether he can perform a certain operation:
++
+```
+/can/{verb}/{resource}
```
pmuir, line 46: Following the Kubernetes model would suggest embedding the verb and resource in the payload, not the path.

_adr/83/index.adoc
pmuir: Yeah, I don't think this introduces any meaningful threats. We don't sufficiently rate limit API calls to mean that by calling this API the actor could get around a rate limit, so I think we aren't exposing anything here here that can't be obtained using brute force.

pmuir (approved):
This looks fine, however I do wonder if we don't just wait until we move to KCP to do it, as the mechanism is mostly present there already.

_adr/83/index.adoc
```diff
@@ -0,0 +1,132 @@
...
+```
++
+Output of the API will be a JSON with the following structure:
++
+```
+{
+  outcome: string,           // enum: granted/denied
```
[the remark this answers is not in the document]
ziccardi: `reason` is optional and "indicates why a request was allowed or denied".
Following kubernetes model I would see it more like `allowed` with a `true` or `false` value

_adr/83/index.adoc
ziccardi: The `SelfSubjectAccessReview` uses a POST method. In our case we don't have to make any change anywhere, so I believe for our API a `GET` method with parameters into the path would suit better.
WDYT?

_adr/83/index.adoc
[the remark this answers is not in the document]
ziccardi: Got it now!

ziccardi:
@pmuir made some changes to get closer to Kubernetes model.

ziccardi:
@emmanuelbernard Mind another look?

_adr/83/index.adoc
```diff
@@ -0,0 +1,132 @@
...
+/can/{verb}/{resource}
+```
++
+Output of the API will be a JSON with the following structure:
++
+```
+{
+  allowed: boolean,
+  metadata: {
+    {key string}:{value}
```
chirino, line 55: Wonder if this could be more structured. Something like:

options: [{
      name: {string},
      values: [{
          value: {any}
          disabled: {string - reason this value is disabled.} 
      }...]
}]
quotas: [{
      name: {string},
      granted: {int},
      used: {int}
}]

_adr/83/index.adoc
chirino: Would you have to implement this for every possible verb/resource combination?  Or will just have a few explicitly listed openapi endpoints that follow this path pattern?

ziccardi: The idea was to have a single endpoint. 
Then a `map` of handlers ([verb-resource][*handler]). If the handler is found into the map, the handler is invoked, otherwise a bad request error is returned.

This way the list of supported 'verb/resource' pair can grow very easily as we need to support new verbs/resources

_adr/83/index.adoc
```diff
@@ -0,0 +1,132 @@
+---
+num: 0 # allocate an id when the draft is created
+title: Permission API
+status: "Draft"
```
tombentley, line 4:
```suggestion
status: "Accepted"
```

_adr/83/index.adoc
```diff
@@ -0,0 +1,132 @@
...
+#applies_padrs: # What PADRs does this ADR apply?
+#applies_patterns: # What APs does this ADR apply?
+---
+## Context and Problem Statement
+Deciding whether a certain operation is allowed can be a complex matter and may require interactions with different components.
+
+Consumers of the Kas Fleet Manager APIs have no way of verifying in advance whether a particular request will be successful.
+The only way to do that, is to perform the same checks that Kas Fleet Manager will do while executing the action.
+
+This brings complexities and duplications. For example, just to check whether the creation of a new kafka is possible or not, at least the following checks must be performed:
```
tombentley, line 18: It's good to be explicit about where those duplications are. IIUC it's in the fleet manager and the API clients (UI and CLI) right?

_adr/83/index.adoc
ziccardi: IMO that is out of the goals of this API. In the case of your example, I believe this API should be used to check if you can create or not a topic (`/can/create/topic`) so that, for example, the UI could enable or disable the button.

Then I believe it is ok to try to create the topic and get an error if a topic with that name already exists.

_adr/83/index.adoc
```diff
@@ -0,0 +1,132 @@
...
+.. Just one quota available: `standard.x2` won't be returned
++
+```
+{
+  allowed: true,
+  metadata: {
+    plans: ['standard.x1'],
+    quota: {
+      granted: 6,
+      used: 5
```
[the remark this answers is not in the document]
ziccardi: The idea was to map it as:
```yaml
type: object
additionalProperties:
  type: string
```

And then document for each `verb/resource` pair what are the keys/values in metadata.

_adr/83/index.adoc
```diff
@@ -0,0 +1,132 @@
...
+.. Terms not accepted
++
+```
+{
+  allowed: false,
+  reason: {
+    code: "ERR-TERMS-01",
+    description: "Required terms have not been accepted",
+    metadata: {
+      acceptance_url: "https://url.to.accept.terms"
```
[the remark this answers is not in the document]
ziccardi: Currently the only KFM error that comes to my mind as one needing to visit an URL is the `terms not accepted` error. That's why I put it into the `metadata` since I was seeing it as an exception

_adr/83/index.adoc
ziccardi: Yes. UI and CLI. Added into the ADR

_adr/83/index.adoc
```diff
@@ -0,0 +1,132 @@
...
+  allowed: boolean,
+  metadata: {
+    {key string}:{value}
+  }
+  reason: {
+    code: string,
+    description: string,
+    metadata: {             // additional details
+      {key string}: {value string}
+    }
```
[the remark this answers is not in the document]
tombentley: In general it's frustrating if you're only told about errors one at a time (e.g. imagine a compiler that bailed on the first error). But it's really a question about whether clients can effectively use info about multiple errors (@pmuir, @wtrocki?). I would advocate for the schema permitting multiple errors, even if we didn't use it immediately. The client can always report just the first one, though we might need to take care about ordering perhaps.

_adr/83/index.adoc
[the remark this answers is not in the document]
tombentley: This is essentially the same point as [Hiram made](https://github.com/bf2fc6cc711aee1a0c2a/architecture/pull/52/files#r843975787), so let's resolve this one and have the discussion there

_adr/83/index.adoc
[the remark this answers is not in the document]
tombentley: Needing to buy more quote was the other example I was thinking of.

wtrocki:
While we still have some feedback and suggestions to apply ADR as whole brings very good pattern that can help us to simplify models for UI and CLI.
From my personal point of view this can be great supplement to our existing API
and replace them completely after we will have confidence for it.

TL;DR - This API will be very useful. Thank you for creating ADR 
as we could later recommend the same pattern for other services as well

_adr/83/index.adoc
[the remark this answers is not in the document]
ziccardi: Oh, I see now what you mean. 
Well, when the KFM invokes the terms API to know if the terms have been accepted or not, the API itself returns the url to be visited to fix the issue. The same does not happens for the AMS API (quota management).

I don't think KFM should store the URL, so I think it would be better to keep it into the `metadata` as a custom attribute.

_adr/83/index.adoc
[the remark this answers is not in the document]
wtrocki: Current errors are blocking entire workflow so returning one should be ok. When returning one we rely on backend to set priorities on which error should be first. For me that is ok but would love to get @riccardo-forina opinion.

As generic pattern/kubernetes I think we need to choose between:

1) Status codes reporting (single error) multiple request

2) Reconciler based status conditions (array)
https://maelvls.dev/kubernetes-conditions/

I do not have much prefference and for CLI and UI it is ok to do 1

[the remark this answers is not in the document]
wtrocki: This can also affect UX - @jgiardino let us know if you have prefference.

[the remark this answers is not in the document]
ziccardi: We could add a `deep=true/false` parameter so that the caller can choose whether they want just one or all of them together. WDYT?

_adr/83/index.adoc
```diff
@@ -0,0 +1,133 @@
+---
+num: 83
+title: Permission API
```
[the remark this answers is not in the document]
ziccardi: Do we all agree about `Option API` ?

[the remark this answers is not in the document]
Rajagopalan-Ranganathan: @pmuir what are your thoughts, naming is always the hardest part :)

_adr/83/index.adoc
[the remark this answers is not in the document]
pmuir: Being able to return multiple errors is a must in my opinion. I agree that right now we only really display one error at a time, but I think we can improe on that.

_adr/83/index.adoc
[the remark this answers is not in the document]
pmuir: Option or CanI are the best I've seen so far.

_adr/83/index.adoc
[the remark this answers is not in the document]
ziccardi: Changed the ADR to include multiple errors

_adr/83/index.adoc
```diff
@@ -0,0 +1,132 @@
...
+}
+```
+
+.. Terms not accepted
++
+```
+{
+  allowed: false,
+  reason: {
+    code: "ERR-TERMS-01",
```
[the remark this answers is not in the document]
ziccardi: My idea was to structure the error codes like this: `ERR-{CATEGORY}-{NUMBER}`
Where ERR is a fixed string, `CATEGORY` is a check category like QUOTA, TERMS, etc, while NUMBER is a progressive number inside the CATEGORY.

_adr/83/index.adoc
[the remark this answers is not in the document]
emmanuelbernard: I thought of `Possibilities API` as well. I find `Can I` clear as well.

_adr/83/index.adoc
```diff
@@ -0,0 +1,156 @@
...
+## Current architecture
+* Kas Fleet Manager currently only allows the ‘try and see if it works’ approach: if what you are trying to do fails, Kas Fleet Manager gives you some details on went wrong, but you have no way of checking in advance if what you want to do is possible
+* API Consumers (like UI) are doing the required checkss into their codebase and every time a change occurs, both Kas Fleet Manager and UI have to be updated to reflect the change
+
+## Proposed architecture
+* We need to create a new API.
+The API will be authenticated and will allow the caller to check in advance whether he can perform a certain operation:
++
+```
+/can/{verb}/{resource}?deep=true
```
tombentley, line 47: Now is this API versioned?

ziccardi: yeah, all our API are versioned. I removed the `/api/v1` prefix for clarity

tombentley: Thanks, I wasn't certain.

ziccardi:
@emmanuelbernard @tombentley @chirino @wtrocki Renamed the ADR and addressed your comments. Mind giving another look?

_adr/83/index.adoc
tombentley: A generic return value for all the checks necessarily constraints this API to only those checks that are common to all resources. That works for the things you're concerned about _right now_ (like terms, entitlements and capacity), but are we sure we're not going to be interested in other checks in the future? For example, capacity depends on the number of streaming units requested (which is a Kafka specific concept, at least right now). So an initial "yes" from the end point might become a "no" once the user has expressed interest in 5SU rather than 1SU. Maybe we don't plan right now to make a 2nd call when they change the number of SU, but are we confident that we won't change our minds about that? By using GET we''d be adopting an API that wasn't capable later refinement to support such finer-grained logic.

Also the generic return type argument is somewhat undermined by also returning arbitrary extra metadata, which clients need to know how to handle in order to make use of.

As a straw-man alternative pattern, I could imagine re-using the resource creation endpoint, only with a different content type for the request which changed the meaning of the request (and the expected request body) to have a dry-run semantic. The request type schema would be changed to be all optional properties, thus allowing the frontend to make requests with partial information. Doing this would tend to keep the logic for the dry-run and the I-really-mean-it cases in the same place, which would be easier to keep in sync, than having a separate endpoint handled by a separate code path. It would also allow us to gracefully change the logic over time to support the finer-grained logic mentioned above.

ziccardi: @tombentley That's not what the api is though for. It is not meant to ask 'Can I create a Kafka with 5SU' and return an answer `yes/no`.
It is meant to answer to `Can I create a Kafka` and answer for example: "Yes, you can create one up to 3SU'.
So I don't think the 'create' endpoint in dryrun mode would fit.
If you think the API this way, the current generic structure will be able to cover this case too.

tombentley: @ziccardi yes, I understand that the straw man is not doing exactly the same thing as is described in this ADR. It is a superset. You can ask "Can I create a Kafka" and get yes/no, and then later during the user interaction you could also ask "Can I create a Kafka in AWS US east" and get yes/no, and later still "Can I create a Kafka of 5 SU in AWS US east" and get yes/no. So it _could be_ more useful, depending on whether the UI would (eventually) make use of this functionality. It is also more strongly typed, since the requests have a schema that's directly related to the schema for instance creation, rather than having to use an arbitrary, service-specific `object` to pass extra information. It is also something that could be implemented gradually, and go as fine-grained as we need.

> So I don't think the 'create' endpoint in dryrun mode would fit.

Since it contains the "Can I create a Kafka" functionality, I don't understand why it doesn't fit. Can you explain?

> the current generic structure will be able to cover this case too.

But it does so at the expense of having a schema for describing the request `metadata` properly.

tombentley: > So it could be more useful, depending on whether the UI would (eventually) make use of this functionality.

@pmuir @jgiardino would/could this actually be of use, do you think?

ziccardi: > Since it contains the "Can I create a Kafka" functionality, I don't understand why it doesn't fit. Can you explain?

Because when you invoke the `create` endpoint, you have to describe exactly what you want to create and the outcome will be about creating exactly the instance you asked (for example, a `standard` instance on AWS US East).

With the `can-I` you ask `can I create a kafka`? and it will return an answer like: "Yes, you can create `standard,eval` on AWS US East or AWS EU, with up to 3SU", so you won't need to make the subsequent calls because it already gives you all the infos about what you can do.

_adr/83/index.adoc
```diff
@@ -0,0 +1,158 @@
...
+  - "Massimiliano Ziccardi"
+tags:
+  - "kafka"
+#applies_padrs: # What PADRs does this ADR apply?
+#applies_patterns: # What APs does this ADR apply?
+---
+## Context and Problem Statement
+Deciding whether a certain operation is allowed can be a complex matter and may require interactions with different components.
+
+This new API will allow the clients to check whether a certain operation is allowed by simply passing the top-level verb (create,delete,list, etc) and receiving whether it is allowed and what are the granted resources.
```
tombentley, line 15:
```suggestion
The clients need to be able to check whether a certain operation is allowed without having all or any of the details about the instance the user wants to create (because the user hasn't completed the form yet).
So the only available inputs are the top-level verb (create, delete, list, etc.) and the resource (e.g. kafka instance)
```

_adr/83/index.adoc
tombentley: Thanks for the clarification @ziccardi, that makes a lot more sense now (and actually one of your examples showed this).

If we had a dedicated endpoint for each resource and verb (rather than a single parameterized endpoint), it would allow us to have a fully typed schema for the request (and possibly response), rather than relying on the untyped-`metadata` property. That would make the API a lot more self-describing and discoverable. Wdyt?

ziccardi:
@tombentley I applied the suggested change as we discussed. I however left the `metadata` for the `reasons` array: that's because we can have many reasons for an action to be refused and each reason will carry different details. 
For example:

- TERMS NO ACCEPTED ERROR: will provide the url to be reached to accept the terms
- QUOTA EXHAUSTED: will return the granted quota, the available quota and the needed quota

_adr/83/index.adoc
```diff
@@ -0,0 +1,155 @@
...
+.. Enquire the quota assigned using the AMS endpoints, loop through the resources and check that enough quota is still available for product RHOSAK and resource RHOSAK
+.. If you have no quota assigned for STANDARD instances (RHOSAK), fallback to EVAL
+.. If you have quota for STANDARD (RHOSAK), but it is 0, fallback to EVAL (RHOSAKTrial)
+.. If you have quota for STANDARD (RHOSAK) and it is greater than 0 but all the granted quota is consumed, the creation should fail
+. Verify that at least one OCD Cluster has capacity for the instance you are going to create
+
+Another point of attention is that currently, any change in the flow requires an update to all the actors/consumers who are carrying out those checks.
+
+## Goals
+* API consumers can query a new KAS Fleet Manager Permission API to check if they can perform a certain action and receive a response with the outcome (allowed/denied) containing all the granted resources or, eventually, the missing requirements.
```
machi1990, line 37:
```suggestion
* API consumers can query a new KAS Fleet Manager Cann-I API to check if they can perform a certain action and receive a response with the outcome (allowed/denied) containing all the granted resources or, eventually, the missing requirements.
```

_adr/83/index.adoc
```diff
@@ -0,0 +1,155 @@
...
+
+## Non-goals
+
+## Current architecture
+* Kas Fleet Manager currently only allows the ‘try and see if it works’ approach: if what you are trying to do fails, Kas Fleet Manager gives you some details on went wrong, but you have no way of checking in advance if what you want to do is possible
```
machi1990, line 42:
```suggestion
* Kas Fleet Manager currently only allows the ‘try and see if it works’ approach: if what you are trying to do fails, Kas Fleet Manager gives you some details on what went wrong, but you have no way of checking in advance if what you want to do is possible
```

_adr/83/index.adoc
```diff
@@ -0,0 +1,155 @@
...
+* API Consumers (like UI) are doing the required checkss into their codebase and every time a change occurs, both Kas Fleet Manager and UI have to be updated to reflect the change
+
+## Proposed architecture
+* We need to create a new API.
+The API will be authenticated and will allow the caller to check in advance whether he can perform a certain operation.
```
machi1990, line 47:
```suggestion
The API will be authenticated and will allow the callers to check in advance whether they can perform a certain operation.
```

_adr/83/index.adoc
```diff
@@ -0,0 +1,155 @@
...
+A different endpoint will be provided for each verb-resource pair, wich, in turn, will provide a schema defining exactly what the answer will look like.
```
machi1990, line 48:
```suggestion
A different endpoint will be provided for each verb-resource pair, which, in turn, will provide a schema defining exactly what the answer will look like.
```

_adr/83/index.adoc
```diff
@@ -0,0 +1,155 @@
...
++
+```
+GET /api/v1/can/{verb}/{resource}?deep=true
+```
++
+The API will stop at the first error unless `deep=true` is specified
```
machi1990, line 54:
```suggestion
The API will stop at the first error unless `deep=true` is specified.
```

_adr/83/index.adoc
```diff
@@ -0,0 +1,155 @@
...
+  allowed: boolean,
+  reasons: [{
+    code: string,
+    description: string,
+    metadata: {             // additional details
+      {key string}: {value string}
+    }
+  }]
+}
+```
```
machi1990, line 68: Is this JSON out for when deep=true? If not, is it possible to have an example of when deep=true?

_adr/83/index.adoc
```diff
@@ -0,0 +1,155 @@
...
+
+.. Terms not accepted
++
+```
+{
+  allowed: false,
+  reasons: [{
+    code: "ERR-TERMS-01",
```
machi1990, line 106: Will it be possible to use the same error codes pattern (see the response output of `/api/kafkas_mgmt/v1/errors` and observe the error code) as the one we current have in kas-fleet-manager? Introducing another form of error code might requires clients having to reason about different error codes patterns which can be inconsistent

machi1990 (commented):
lgtm

I've left some comments/suggestions/questions

_adr/83/index.adoc
```diff
@@ -0,0 +1,155 @@
...
+
+## Alternatives Considered / Rejected
+
+* Using Authorino. This approach has been discarded because Authorino is more suited for a 'granted/denied' answer than for querying for permissions and producing complex responses
+
+## Challenges
+* Client application will have to be updated to take advantage of the new API
+
+## Dependencies
+* Current check logic implemented in the Kas Fleet Manager needs to be moved to a common, reusable framework that will be used by both the Kas Fleet Manager middlewares and the new Permission API
```
machi1990, line 149:
```suggestion
* Current check logic implemented in the Kas Fleet Manager needs to be moved to a common, reusable framework that will be used by both the Kas Fleet Manager middlewares and the new Can-I API
```

_adr/83/index.adoc
jgiardino: Updates were made since this thread started, so I think this is the relevant update to this conversation:
```
{
  allowed: boolean,
  reasons: [{
    code: string,
    description: string,
    metadata: {             // additional details
      {key string}: {value string}
    }
  }]
}
```
and then is this an example of what the `metadata` could include?
```
{
  allowed: true,
  plans: ['standard.x1'],
  quota: {
    granted: 6,
    used: 5
  }
}
```

In the context of the UI, knowing what's `granted` vs `used` is definitely helpful. 
Knowing what regions are unavailable is also helpful.
It would be great to know if a region has limited capacity, so if a region can support another instance that is size 1, but not size 2, I'd like to communicate that as shown in [this example](https://docs.google.com/presentation/d/1U4j27K3L-9Ry0GPe-fcKoU48dCPlEJsfoXJUpldAY3U/edit?pli=1#slide=id.g118b16caf36_0_83):
image

_adr/83/index.adoc
ziccardi: This is the output both when deep=true and deep=false.
The difference is that when deep=true, the `reasons` array will always have a size of 1

_adr/83/index.adoc
ziccardi: Sure.

_adr/83/index.adoc
machi1990: Thanks Massimo. Is it possible for you to share an example of both when deep=false and deep=true to make sure I really get the difference? Thanks

_adr/83/index.adoc
```diff
@@ -34,24 +34,24 @@ currently UI and CLI have to perform at least the following checks:
...
-* API Consumers (like UI) are doing the required checkss into their codebase and every time a change occurs, both Kas Fleet Manager and UI have to be updated to reflect the change
+* Kas Fleet Manager currently only allows the ‘try and see if it works’ approach: if what you are trying to do fails, Kas Fleet Manager gives you some details on what went wrong, but you have no way of checking in advance if what you want to do is possible
+* API Consumers (like UI) are doing the required checks into their codebase and every time a change occurs, both Kas Fleet Manager and UI have to be updated to reflect the change
 
 ## Proposed architecture
 * We need to create a new API.
-The API will be authenticated and will allow the caller to check in advance whether he can perform a certain operation.
-A different endpoint will be provided for each verb-resource pair, wich, in turn, will provide a schema defining exactly what the answer will look like.
+The API will be authenticated and will allow the callers to check in advance whether he can perform a certain operation.
```
machi1990, line 47:
```suggestion
The API will be authenticated and will allow the callers to check in advance whether they can perform a certain operation.
```

_adr/83/index.adoc
ziccardi: here: https://github.com/bf2fc6cc711aee1a0c2a/architecture/pull/52/files#diff-01aa0bf716b701e03b5bacdf09f7b0deab40d658c388fa3e272db1b6255f5cd5R115

machi1990: Perfect, thanks. I somehow missed that. 

So when `deep=false`, the same example might look like
```
{
  allowed: false,
  reasons: [{
    code: "ERR-TERMS-01",
    description: "Required terms have not been accepted",
    metadata: {
      acceptance_url: "https://url.to.accept.terms"
    }}
  ]
}
```
because we stopped at the first error which was encountered, is my understanding correct?

_adr/83/index.adoc
ziccardi: @jgiardino 
```
{
  allowed: boolean,
  reasons: [{
    code: string,
    description: string,
    metadata: {             // additional details
      {key string}: {value string}
    }
  }]
}
```
is what ALL new endpoints (verb/resources pair) will return. Then each of them will have its own schema adding other details.
For example, the `create/kafka` will add the `plans` and `quota`.
The `metadata` is only in the errors. See [here](https://github.com/bf2fc6cc711aee1a0c2a/architecture/pull/52/files#diff-01aa0bf716b701e03b5bacdf09f7b0deab40d658c388fa3e272db1b6255f5cd5R115) for an example.

> It would be great to know if a region has limited capacity, so if a region can support another instance that is size 1, but not size 2, I'd like to communicate that as shown in [this example](https://docs.google.com/presentation/d/1U4j27K3L-9Ry0GPe-fcKoU48dCPlEJsfoXJUpldAY3U/edit?pli=1#slide=id.g118b16caf36_0_83):

Totally agree.

_adr/83/index.adoc
ziccardi: Exactly

machi1990: Thank you.

tombentley (approved):
LGTM, many thanks @ziccardi!

wtrocki (approved):

_adr/83/index.adoc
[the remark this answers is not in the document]
jgiardino: So the benefit of providing a URL is so that both the UI and CLI have this information, correct?

Regarding the use case @tombentley mentioned, we do have a Kafka overview page — https://console.redhat.com/application-services/streams/overview — that includes CTAs for getting or updating a subscription. Eventually, pricing information will live on redhat.com, so the design will change when that happens. Maybe we would have a single CTA on the Kafka overview page related to getting or updating a subscription that points to the page on redhat.com 🤷‍♀️ I like the idea of providing this URL, as Tom suggests, but I don't know exactly how these pages and contents will evolve.

_adr/83/index.adoc
jgiardino: Thanks for the clarification.

emmanuelbernard (approved):
lgtm

PR #59, opened 2022-05-25
Add ADR-79 in Accepted state

miguelsorianod:
ADR-79: Authorization for Control Planes

This ADR proposed a way to extract Authorization-related code from the Fleet Manager and perform Authorization by using [Authorino](https://github.com/kuadrant/authorino) as an [Envoy’s External Authorization Filter/Server](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_authz_filter). As such, Authorino would become the Authorization layer for Control Plane services that desire Authorization.

### DIFF
--- a/_adr/79/index.asciidoc
+++ b/_adr/79/index.asciidoc
@@ -1,0 +1,11 @@
+---
+num: 79
+title: "Authorization for Control Planes"
+status: "Accepted"
+authors:
+  - "Miguel Soriano"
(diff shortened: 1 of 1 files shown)


miguelsorianod:
> Not sure linking to nthe google docs is the way for new ADRs, added to the meeting for conversation

I agree. But this ADR is not a new one. "old" ADRs where we were still writing them in gdocs are linked.

PR #62, opened 2022-05-27
Add templates for AP and PADRs

tombentley:
This PR adds templates (with number 0) for APs and PADRs, like we already have for ADRs.

### DIFF
--- a/_ap/0/index.adoc
+++ b/_ap/0/index.adoc
@@ -18,6 +18,7 @@ tags:
 
 ## Intent
 // Summarize in a single sentence what the pattern tries to achieve.
+//
 // For example:
 //
 // Define a policy for exposing metrics to customers in a consistent way.
@@ -25,9 +26,10 @@ tags:
 ## Motivation
 
 // In a few paragraphs describe the motivating factors for this pattern
+//
 // For example:
 // 
-// Prometheus scrape endpoints are the defacto standard for exposing metric information for collection
+// Prometheus scrape endpoints are the defacto standard for exposing metric information for collection.
 // However this format and the associated conventions are not sufficient in themselves to ensure
 // that metrics are exposed in a consistent way. Problems include:
 // * Insufficient guidance on metric and label naming to provide consistency and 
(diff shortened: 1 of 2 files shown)


tombentley:
@emmanuelbernard PTAL. The PADR one is basically identical to the ADR one, so I assume that's not contentious. The AP one is a little more interesting.

_ap/0/index.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+// Top style tips:
+// * Use one sentence per line
+// * No unexpanded acronyms
+// * No undefined jargon
+
+// You don't have to use the following sections, but they provide a 
+// useful structure for writing a clear document.
+
+## Intent
```
emmanuelbernard, line 19: Reading these sections I have a harder time to project myself.
Can you provide a fictitious short example of an AP and apply the template?
This might even be part of the template to help less abstract people get onboard.

Another data point I'd like is some how who and how many services use that pattern to know when to decide to move to a PADR.

tombentley: > Can you provide a fictitious short example of an AP and apply the template?

@emmanuelbernard I've added an on-going example based on the metrics API which @JameelB is working on. Is that the sort of thing you were looking?

> Another data point I'd like is some how who and how many services use that pattern to know when to decide to move to a PADR.

That's not really something the AP itself is intended to address. IIRC we said that each time a service decided to follow an AP there should be a trivial ADR (with an `applies_ap` in the frontmatter) to record that fact ("Apply AP-X to Service Y"). But that in itself doesn't give us a number of times the AP has been used. Do you think we need that, or could we expect it to be caught during the review of the ADR?

_ap/0/index.adoc
```diff
@@ -0,0 +1,68 @@
...
+// the metrics are exposed
+
+## Consequences
+
+// Explain any consequences of using this pattern
+//
+// For this metrics example consequences would include:
+// * A mechanism for API evolution
+// * Consistency in naming of metrics and labels within this service
+// * Consistency with other services which also apply this AP
```
emmanuelbernard, line 68: Add maybe a negative consequence example

_ap/0/index.adoc
```diff
@@ -0,0 +1,68 @@
...
+## Applicability
+
+// Call out when this pattern might apply (and when it should not apply, if relevant)
+// For example:
+//
+// This pattern applies to any service which exposes metrics as Prometheus-scrape endpoints.
+// It does not apply to services which provide access to metrics in other formats, or via
+// means other than an API (e.g. metric visualization via a system such as Grafana).
+
+## Structure
```
emmanuelbernard, line 47: I like all sections exceit this one for which the title is not speaking to me.
It feels like the core of the pattern. Propositions:

* What the pattern is
* Description
* Detail

Right now, *Description* is what speaks to me the most.

emmanuelbernard (commented):
Minor clean up before approval

tombentley:
I made a few updates. @emmanuelbernard please take another look.

Issue #61, opened 2022-05-26
Create customer exposed metric naming policy


JameelB:
The current customer exposed metrics have inconsistent naming. There are currently two types of name format:
- recording rules (i.e. `kafka_topic:kafka_log_log_size:sum`)
   - This follows the format of \<level\>:\<metric name\>:\<aggregation\>
   - This makes sense when you know how it's implemented but it's an implementation detail.
- raw metrics (i.e. `kafka_controller_kafkacontroller_offline_partitions_count`).
   - This is taken directly from kafka metrics

We need to find a consistent logical naming scheme, bearing in mind that these metric names are an API and it's possible that the implementation of some of them may evolve over time.

We should create an ADR to define this policy. It should cover the following topics:
- How do we request new metrics?
- How do we name new metrics? 
- How would we deal with existing customer exposed metrics already available in our API?
- How can we version existing/new metrics?
- Competitive analysis, how are others doing this?

Related Epic: https://issues.redhat.com/browse/MGDSTRM-7080

tombentley:
I can do this...

tombentley:
I think this is really a platform wide concern. I.e. in general any managed service might expose a prometheus-scrapable endpoint to customers, and when doing so is creating a public API contract which therefore needs managing. Furthermore this could be handled in a consistent way across services. So I wonder if this should really be a AP, with a brief ADR which says "We're applying this pattern to the Kafka service". Wdyt @emmanuelbernard, @pmuir ?

tombentley:
@JameelB I've added/reserved ADR-87 for now.

JameelB:
Thank you @tombentley!

emmanuelbernard:
Yes let’s have an AP

JameelB:
@tombentley @emmanuelbernard created this PR to reserve ap-16 for this issue: https://github.com/bf2fc6cc711aee1a0c2a/architecture/pull/65

PR #64, opened 2022-06-01
adr-87: customer facing metrics naming policy

JameelB:
ADR-87: Customer Facing Metrics Naming Policy

This ADR proposes a naming policy that can be applied to all of our customer facing metrics so that they are consistent and clear to our end users.

See related issue: https://github.com/bf2fc6cc711aee1a0c2a/architecture/issues/61

### DIFF
--- a/_adr/87/index.adoc
+++ b/_adr/87/index.adoc
@@ -1,13 +1,13 @@
 ---
 num: 87 
-title: Prometheus endpoint APIs 
-status: "Draft" # One of Draft, Accepted, Rejected
+title: Exposing metrics via the KAS Fleet Manager API
+status: "Draft"
 authors:
-  - "JameelB" # One item for each author, as github id or "firstname lastname"
+  - "JameelB"
 tags:
-  - "kafka" # e.g. kafka, connectors, registry
+  - "kafka"
 applies_padrs: # What PADRs does this ADR apply?
-applies_patterns: # What APs does this ADR apply?
+applies_patterns: 16
 ---
 
 // Top style tips:
@@ -15,34 +15,214 @@ applies_patterns: # What APs does this ADR apply?
 // * No unexpanded acronyms
 // * No undefined jargon
 
-// No need for a title heading, it's added by the template
-
 ## Context and Problem Statement
-// What is the background against which this decision is being taken?
+The KAS Fleet Manager API provides the following endpoints which end users can use to collect metrics of their own Kafka instances:
+
+* https://api.openshift.com/?urls.primaryName=kafka%20service%20fleet%20manager%20service#/default/getMetricsByInstantQuery[/api/kafkas_mgmt/v1/kafkas/{id}/metrics/query]: returns the most recent values for each metric associated with the specified Kafka.
+This returns metrics in a JSON format.
+* https://api.openshift.com/?urls.primaryName=kafka%20service%20fleet%20manager%20service#/default/getMetricsByRangeQuery[/api/kafkas_mgmt/v1/kafkas/{id}/metrics/query_range]: returns values for each metric associated with the specified Kafka over a time range specified in the request.
+This returns metrics in a JSON format.
+* https://api.openshift.com/?urls.primaryName=kafka%20service%20fleet%20manager%20service#/default/federateMetrics[/api/kafkas_mgmt/v1/kafkas/{id}/metrics/federate]: returns the most recent values for each metric associated with the specified Kafka.
+This returns metrics in a Prometheus Text format and is scrapable by any monitoring systems that accepts this format.
+
+Each of these endpoints returns the exact same metrics.
+The only difference between these is the format of the response.
+
+The metrics returned in these endpoints are what we call customer facing metrics.
+All available customer facing metrics are documented in the app-services-guides: https://github.com/redhat-developer/app-services-guides/tree/main/docs/kafka/metrics-monitoring-kafka[Monitoring metrics in OpenShift Streams for Apache Kafka]. 
+
+We currently expose two different types of customer facing metrics:
+
+* Raw Kafka metrics -- exposed by Kafka instances.
+* Recording rules -- configured by us for any Kafka metrics requiring pre-computed aggregation.
+
+This results in multiple problems:
+
+* *Consistency*: These two types of metrics follow different naming schemes, resulting in inconsistent metric names.
+* *Repetition*: The raw Kafka metrics contain a lot of repetition i.e. `kafka_controller_kafkacontroller...`.
+* *Exposure of implementation details*: The name of the recording rules metrics exposes implementation details i.e. `kafka_namespace:haproxy_server_bytes_in_total:rate5m`, end users do not need to know about namespaces and OpenShift.
+
+Since we don't have a formal process and naming policy for adding customer facing metrics, we often reach a situation where the resulting metric names are confusing and misleading (i.e. https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/issues/688[misleading metric name for kafka topic partitions count]).
+
 
 ## Goals
-// Bulleted list of outcomes that this ADR, if accepted, should help achieve
+- Define a naming policy for existing and new customer facing metrics.
+- Define a process for requesting new customer facing metrics to the API.
+- Define a process for migrating the existing customer facing metrics to follow the new naming policy.
+
 
 ## Non-goals
-// Bulleted list of outcomes that this ADR is not trying to achieve.
+- Metrics API versioning
++
+The metrics API is currently part of the KAS Fleet Manager API.
+We need to consider KAS Fleet Manager API as a whole in order to have consistent versioning strategy for all of our endpoints, including customer facing metrics and the metrics API.
++
+
++
+This also includes different deprecation periods with different lifecycle tags and the idea of separating the metrics API from the rest of the KAS Fleet Manager API to have isolated versioning.
+This is out of scope of this ADR and should have its own separate discussion, epic or ADR.
++
+- Parameterize metric aggregation
+- Provide endpoint for metrics discovery
+
 
 ## Current situation
-// Where are we now?
+### Requesting new customer facing metrics
+There is currently no formal process that we follow when a new metric is requested to be added to the API. 
+
+The request is handled by the Control Plane team to add the metrics to the KAS Fleet Manager API as well as adding new recording rules to the Observability configuration if needed.
+However, the Kafka Integrations team would have the knowledge of what metrics to use and how it should be aggregated, if required.
+All discussions around which metric to use and how its aggregated is done via ad-hoc chat conversations, in the JIRA issue or the Github pull request.
+Discussion to reach a formal agreement cross teams for the metric name is not done.
+
+
+### Naming scheme
+The names of the current customer facing metrics are based on the following:
+
+* Raw Kafka metrics: The names of these metrics are not changed, they are taken directly from Kafka metrics. These have the following name format:
+** For example: `kafka_controller_kafkacontroller_offline_partitions_count`
+
+* Recording rules: We configure these and name them based on the following format `<level>:<metric name>:<aggregation>`
+** For example: `kafka_topic:kafka_log_log_size:sum`
+** This format is used for all of our recording rules, most are not exposed to end users.
+** This makes sense when you know how it's implemented but this exposes implementation details to end users.
+** A list of the customer facing recording rules can be seen in the https://github.com/bf2fc6cc711aee1a0c2a/observability-resources-mk/blob/main/resources/prometheus/user-facing-recording-rules.yaml[observability-resources-mk] configuration repository.
+
+
+### Exposing new metrics in the KAS Fleet Manager API
+All of the customer facing metrics originate from the data plane cluster in which the Kafka instance is deployed in.
+A Prometheus instance is deployed in each of the data plane cluster which uses https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write[remote-write] to push these metrics to Observatorium.
+This acts as the central repository for all metrics, included in the remote-write configuration, from each data plane cluster.
+A list of the metrics included in the remote-write configuration can be found in the https://github.com/bf2fc6cc711aee1a0c2a/observability-resources-mk/blob/main/resources/prometheus/remote-write.yaml[Observability resources] repository.
+
+KAS Fleet Manager acts as a proxy to Observatorium.
+It calls the Observatorium _/query_ and _/query_range_ endpoints to retrieve metrics we consider as 'customer facing'.
+A list of the metrics retrieved by KAS Fleet Manager can be found in the https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/client/observatorium/api.go[observatorium] client within the project. When new metrics are added to this list, they become available in the KAS Fleet Manager _/metrics/query_ and _/metrics/query_range_ endpoints. In order to make new metrics available to the _/metrics/federate_ endpoint, their metadata must be added to this https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/kafka/constants/metrics.go[file] in KAS Fleet Manager.
+
+KAS Fleet Manager controls the labels exposed for each metric returned in all endpoints. 
+Metrics retrieved from Observatorium are also filtered so that end users will only receive metrics for their own Kafka instance.
+Apart from these, no additional aggregation, filtering or modification of the metric is done by KAS Fleet Manager.
 
 ## Proposal
-// What is the decision being proposed
+### Metric and Label Naming
+All metric and label names must adhere to the policy stated in link:../../_ap/16/index.adoc[AP-16: Naming Policy for Customer Facing Metrics].
 
-### Threat model
-// Provide a link to the relevant threat model. 
-// You must either update an existing threat model(s) to cover the changes made by this ADR, or add a new threat model.
+### Kafka Metric Categories
+The customer facing Kafka metrics can be divided into multiple categories.
+Categorizing these metrics into logical groups can help us formulate the metric name and document it.
+
+|===
+|Category | Description | Metric Prefix
+
+|Instance
+|General Kafka metrics not specific to any other categories.
+|`kas_instance_`
+
+|Broker
+|Metrics associated to Kafka brokers
+|`kas_broker_`
+
+|Topic
+|Metrics associated to Kafka topics
+|`kas_topic_` (Even though the topic metrics already have a 'topic' label, including this improves the clarity of the metric name)
+|===
+
+CAUTION: When adding new metric categories to the service, avoid using categories that exposes the implementation details of the service.
+Metric names are part of the API contract and will be documented.
+For example, categories like `topic` and `broker` are ok to use as they are foundational concepts in Kafka and are intrinsic to the service. 
+However, components where these metrics originated from, e.g. canary or kafka-exporter, are implementation details.
+These should not be used as a metric category.
+
+#### Labels
+At minimum, each metric should have a label that identifies the Kafka instance if possible.
+This would be useful if a user has multiple Kafka instances.
+If they want to integrate all of their Kafka metrics into one monitoring system, the label(s) can act as an identifier.
+
+### Process for requesting new customer facing metric
+When requesting for a new metric to be exposed in the KAS Fleet Manager API, a new issue in JIRA or Github must be created.
+The description of this issue should have sufficient information on what this new metric should provide in order to identify the metric to be used and how it should be documented in later stages.
+
+The issue created must have the following sub-tasks and they must be completed in order:
+
+. Identify the metric and labels to be used and how it will be aggregated if required.
+** This should be assigned to the team that has the knowledge of the requested metric.
+** [.underline]#Definition of Done#:
+*** Metric to be used is identified.
+*** (Optional) Aggregation of the metric determined.
+*** Category which this new metric belongs to is identified.
+**** See link:#kafka-metric-categories[Kafka Metric Categories] to view existing metric categories and notes on adding new ones to the service.
+*** Cardinality calculated
+*** New metric approved by the Running the Services team to be included in the remote-write to Observatorium.
+
+. Agreement on metric name and labels to be exposed
+** This should be assigned to the Control Plane team.
+** [.underline]#Definition of Done#:
+*** Labels to be exposed are identified.
+*** Name of the metric and labels exposed are determined.
+This should adhere to the metric and label naming policy as described above.
+*** The metric and exposed label names must be approved by the following people/teams:
+**** Requester
+**** Control Plane team
+**** Kafka Integrations team
+
+. Expose the metric in the KAS Fleet Manager API
+** This should be assigned to the Control Plane team.
+** Any tasks for adding the new metric to the API should be done in this stage.
+This includes adding the new metric to the Prometheus remote-write configuration and any code changes to the KAS Fleet Manager repository.
+** [.underline]#Definition of Done#:
+*** New customer facing metric should be included in the response of the following KAS Fleet Manager API endpoints:
+**** https://api.openshift.com/?urls.primaryName=kafka%20service%20fleet%20manager%20service#/default/getMetricsByInstantQuery[/api/kafkas_mgmt/v1/kafkas/{id}/metrics/query]
+**** https://api.openshift.com/?urls.primaryName=kafka%20service%20fleet%20manager%20service#/default/getMetricsByRangeQuery[/api/kafkas_mgmt/v1/kafkas/{id}/metrics/query_range]
+**** https://api.openshift.com/?urls.primaryName=kafka%20service%20fleet%20manager%20service#/default/federateMetrics[/api/kafkas_mgmt/v1/kafkas/{id}/metrics/federate]
+
+. Document the new metric
+** This should be assigned to the Documentation/Customer Content Support team.
+** [.underline]#Definition of Done#:
+*** New metric documented in the app-services-guides: https://github.com/redhat-developer/app-services-guides/tree/main/docs/kafka/metrics-monitoring-kafka[Monitoring metrics in OpenShift Streams for Apache Kafka].
+
+
+### Migrating existing metrics to the new naming policy
+KAS Fleet Manager does not support API versioning at the moment so we would need to support both existing and new metrics during the migration process.
+
+The Kafka console provides dashboards which can display data up to the last 7 days (see the design in https://redhat-developer.github.io/app-services-ui-components/?path=/story/kafka-metrics--all-ready[Storybook]).
+Since the revised metric names will be added as new metrics, they will have no history.
+However, there should be sufficient time between adding the metrics to KAS Fleet Manager and the UI team to start working on switching to the new metrics so that there would be no problem with the metric history.  
+Because of this we should be able to expose the new metrics in KAS Fleet Manager and mark the existing metrics as deprecated as soon as they're configured to be pushed to Observatorium. 
+
+The existing metrics shall be marked as deprecated as per the MAS API deprecation policy.
+The deprecation will be communicated as follows:
+
+* The KAS Fleet Manager OpenAPI specification will document the deprecated metrics and their alternatives in each of the endpoint descriptions.
+* The metrics listed in the https://github.com/redhat-developer/app-services-guides/tree/main/docs/kafka/metrics-monitoring-kafka[Monitoring metrics in OpenShift Streams for Apache Kafka] documentation shall be marked as 'deprecated', pointing the users to use the newly revised metrics.
+* An email will be sent out to mas-devel and mas-all mailing list as per the deprecation policy.
+* Deprecation period shall be 60 days.
+** This was agreed upon as there is currently not a lot of users using these metrics apart from the UI.
+Note that in the future, this may be based on the different deprecation periods agreed upon for individual API lifecycle tags (out of scope in this ADR) or based on usage analysis of the deprecated metric.
+** After this time has passed, the deprecated metrics shall be removed from the API and documentation.
+
+
+## Threat model
+Not Applicable.
+
 
 ## Alternatives Considered / Rejected
+* Renaming the metric and label names in the KAS Fleet Manager. 
+KAS Fleet Manager should only act as a proxy to Observatorium, it shouldn't do any additional filtering/aggregation.
 
 ## Challenges
 // What are the costs/drawbacks of the proposed decision?
+* Communicating the change to our consumers
+* Migration of existing metrics to the new naming policy.
+Since we will be following the deprecation policy, we will keep both existing and new metrics until the deprecation period is over.
+This results in duplicated metrics pushed to Observatorium during this period.
+
 
 ## Dependencies
 // What are the knock-on effects if this decision is accepted?
+* UI will need to migrate to the new metrics for the dashboards in the Kafka console.
+* Any other clients using the metrics API will need to migrate to the new metrics.
+
 
 ## Consequences if not completed
 // What are the knock-on effects if this decision is not accepted?
+* Metric name and labels will continue to be inconsistent, repetitive, misleading, confusing and expose implementation details.
--- a/_ap/16/index.adoc
+++ b/_ap/16/index.adoc
@@ -3,7 +3,7 @@ num: 16
 title: "Naming policy for customer facing metrics"
 status: "Draft"
 authors: 
-  - "JameelB" # One item for each author, as github id or "firstname lastname"
+  - "JameelB"
 tags:
   - "kafka" # e.g. kafka, connectors, registry
 ---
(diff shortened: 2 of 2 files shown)


_adr/87/index.adoc
```diff
@@ -15,31 +15,146 @@ applies_patterns: # What APs does this ADR apply?
 // * No unexpanded acronyms
 // * No undefined jargon
 
-// No need for a title heading, it's added by the template
-
 ## Context and Problem Statement
-// What is the background against which this decision is being taken?
+The KAS Fleet Manager API provides endpoints which end users can use to collect metrics of their own Kafka instances.
+The metrics returned in these endpoints are what we call customer facing metrics.
```
tombentley, line 20: It's worth saying that:
* these endpoints are intended to be scraped by Prometheus. I.e. they're not just any old REST API which we can evolve _at will_.
* the Prometheus format doesn't have any built in mechanisms for handling things like versioning or deprecation, and therefore this policy should address these things.
* the Prometheus format does come with guidelines on metric naming, which this policy needs to explicitly either accept, or reject (with reasons for why).

tombentley (commented):
@JameelB this is a good start. 

What competitors are doing is certainly interesting, and might provide some ideas/inspiration, but I don't think we need so much detail in this document. More important is to think about what makes sense for _our_ service; how it might evolve in the future and what guidelines we need to ensure we can achieve a consistent and predictable API over time. Here are some thoughts:

* What we come up with here will need to be compatible with the Architectural Pattern, so we should be mindful of, for example, finding naming schemes that will work for things other than just Kafka.
* The Kafka service will (inevitably) become more complex over time; there might be the need to expose metrics from components that we currently think of as being behind-the-scenes, or an implementation detail. So probably we need guidelines to avoid things like project names (e.g. strimzi, cruise control, kafka exporter) from appearing in metric names. 
* Similarly, for recording rules, how important it is _to the customer_ to know the aggregation period? If so, should it be part of the metric name, or is is sufficient to have it in the documentation?
* We probably need discussion about high cardinality labels: 
     * What counts as high-cardinality?
     * When is it OK to publish such metrics (i.e. we move the problem of high cardinality labels to the customer)?
     * Any advice on how to avoid it?
* How are we going to address the later need for change? e.g. are we going to assert that metrics must be published through an endpoint which contains a version number in the path?

JameelB:
Thanks for the feedback @tombentley!

@emmanuelbernard also agreed on having an architecture pattern for this so I was thinking of adding those details you mentioned above for the naming policy in that document then linking that in the `Proposal` section of this adr. wdyt?

_adr/87/index.adoc
JameelB: > these endpoints are intended to be scraped by Prometheus. I.e. they're not just any old REST API which we can evolve at will.

This is not necessarily true. We do have the `/metrics/query` and `/metrics/query_range` which are used for the Kafka dashboards in the UI and these returns metrics in a JSON format. These endpoints, I think we can evolve at will. The only endpoint we have that's scrapable by Prometheus is `/metrics/federate` and that needs to conform to a Prometheus Text format. I'll make that more clear in this document as well. 

> What competitors are doing is certainly interesting, and might provide some ideas/inspiration, but I don't think we need so much detail in this document.

It's the first time I've done competitive analysis with an ADR so I wasn't sure how much details I needed to put into that. Do you reckon 1/2 examples would've been sufficient?

JameelB: Thanks, that makes sense. Since others were interested in those results, it might be best to record that in the epic/jira issue instead :thinking: 

I'll remove that section from this document :)

emmanuelbernard:
@JameelB you mean the AP points to a section of the ADR?
I think I would prefer the other way around intuitively, no strong arguments.
To be honest, I think having a full copy is easier to consume.

JameelB:
@emmanuelbernard I meant the other way around. This ADR point to the AP. The AP will have details/structure of the naming policy. Basing this on @tombentley's comment on this [issue](https://github.com/bf2fc6cc711aee1a0c2a/architecture/issues/61#issuecomment-1139431155).

_adr/87/index.adoc
```diff
@@ -15,31 +15,92 @@ applies_patterns: # What APs does this ADR apply?
...
 
 ## Goals
-// Bulleted list of outcomes that this ADR, if accepted, should help achieve
+- Define a naming policy for existing and new customer facing metrics.
+- Define a process for requesting new customer facing metrics to the API.
+- Define a process for migrating the existing customer facing metrics to follow the new naming policy.
 
 ## Non-goals
-// Bulleted list of outcomes that this ADR is not trying to achieve.
+- Propose solution for metrics API versioning.
```
JameelB, line 53: After thinking about it, it may be best for us not to include this in the ADR but take it into consideration as part of the whole KAS Fleet Manager API versioning strategy. This has also been brought up a lot in recent discussions within the Control Plane team.

_ap/16/index.adoc
```diff
@@ -1,9 +1,9 @@
 ---
-num: 0
+num: 16
```
JameelB, line 2: Let me know if this should be separated into a different PR @tombentley

_ap/16/index.adoc
```diff
@@ -13,56 +13,131 @@ tags:
...
+*<service-name>_<optional:sub-component>_<descriptor>_<optional:base-unit>_<optional:type|aggregate>*
 
-// What components and/or teams are directly involved
-//
-// For this metrics example this would include the customer and the control plane via which
-// the metrics are exposed
+* *service-name*: The metric name should be prefixed with the service name (e.g. `rhosak`)
+* *sub-component* (optional): The sub-component of the service that the metric measures.
+This should be used for logically categorizing the metrics. 
+The different sub-components of the service must be agreed on by the team (e.g. `server`, `producer`, `consumer`).
```
JameelB, line 48: The example sub-components are based on the Kafka metric categories. Will change these to the sub-components that we agree on for rhosak.

_ap/16/index.adoc
```diff
@@ -13,56 +13,131 @@ tags:
...
+* Expose underlying components in the metric or label name.
+For example, avoid including project names like 'strimzi' or 'kafka-exporter' or OpenShift/Kubernetes resources like 'namespaces' or 'projects'.
+
+### Deprecating Metrics
+The team's API deprecation policy must be followed when deprecating metrics. 
+
+The length of the deprecation period should be agreed upon by the team and applied for all metrics exposed in their API.
+
+
+### Examples
```
JameelB, line 86: Similar to [https://github.com/bf2fc6cc711aee1a0c2a/architecture/pull/64/files#r892526338](comment) above. Examples here were based on sub-components `server`, `producer`, `consumer`. Will change these once we agree what sub-components we'll have for rhosak.

_adr/87/index.adoc
```diff
@@ -1,48 +1,230 @@
...
+Once 7 days has passed, since the new metrics has been added, the existing and new metrics would have the same history for the UI.
+At this point, we can expose the new metrics in KAS Fleet Manager and mark the existing metrics as deprecated. 
+
+The existing metrics shall be marked as deprecated as per the MAS API deprecation policy.
+The deprecation will be communicated as follows:
+
+* The KAS Fleet Manager OpenAPI specification will document the deprecated metrics and their alternatives in each of the endpoint descriptions.
+* The metrics listed in the https://github.com/redhat-developer/app-services-guides/tree/main/docs/kafka/metrics-monitoring-kafka[Monitoring metrics in OpenShift Streams for Apache Kafka] documentation shall be marked as 'deprecated', pointing the users to use the newly revised metrics.
+* An email will be sent out to mas-devel and mas-all mailing list as per the deprecation policy.
+* Deprecation period shall be 60 days.
```
JameelB, line 185: Subject to change

_adr/87/index.adoc
```diff
@@ -1,48 +1,232 @@
...
+|None
+
+|Broker
+|Metrics associated to Kafka brokers
+|broker
+
+|Topic
+|Metrics associated to Kafka topics
+|None (the topic metrics already have a 'topic' label so we should avoid using this in the metric name, as per the naming policy)
+|===
```
JameelB, line 119: Based on the existing categories we have in our [documentation](Based on the existing categories in our documentation ). Subject to change

_adr/87/index.adoc
```diff
@@ -1,48 +1,232 @@
...
+
+Examples of sub-component usage in the metric name:
+
+* For *Cluster* metrics, the metric prefix would be `rhosak_`.
+* For *Broker* metrics, the metric prefix would be `rhosak_broker_`.
+* For *Topic* metrics, the metric prefix would be `rhosak_`.
+
+### Converted Metrics
+<TODO>
```
JameelB, line 128: Will add once we've agreed on the categories/sub-components.

_adr/87/index.adoc
```diff
@@ -1,48 +1,442 @@
...
+At minimum, each metric should have an `instance_name` label that identifies the Kafka instance.
+This would be useful if a user has multiple Kafka instances.
+If they want to integrate all of their Kafka metrics into one monitoring system, the `instance_name` can act as an identifier.
+
+#### Cluster Metrics
+|===
+|Old Metric Name | Revised Metric Name | Type | Old Labels | Revised Labels | Description
+
+|kafka_instance_max_message_size_limit
+|rhosak_message_size_limit_bytes
```
machi1990, line 137: isn't `rhosak` product specific? I am thinking of upstream / product and someone in the external might get confused if they see a metric with that name.

JameelB: I think this is specific to our product though. Unless there is a plan to make the [observability-resources-mk](https://github.com/bf2fc6cc711aee1a0c2a/observability-resources-mk) configuration public :thinking: I don't remember any discussion around the plan/alternative for that yet?

machi1990: There won't be any rename happening on the way KFM retrieve the metrics?

machi1990: > no there shouldn't be, KFM should only act as a proxy. It shouldn't rename any metric/its labels.

+1. This is correct.

What I was saying is that KFM will then have to start requesting for the metrics using the revised name. 
Instead of `kafka_controller_kafkacontroller_global_partition_count`, it'll use `rhosak_broker_global_partition_total`.

JameelB: ah good point! we would also need to change it in the metrics metadata: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/kafka/constants/metrics.go

We could make the prefix configurable if we want to have separate names between upstream/downstream. The upstream/downstream observability-resources-mk config would have different names for the user facing metrics. Do we have an upstream name yet? :thinking: 

Otherwise, we can have a generic name for both that's not product specific. Do you have anything in mind? :slightly_smiling_face:

JameelB:
This is ready for initial review. Please take a look when you get a chance @pb82 @tombentley @k-wall @StevenTobin @riccardo-forina

_adr/87/index.adoc
machi1990: > We could make the prefix configurable if we want to have separate names between upstream/downstream.

Yes. 

> We could make the prefix configurable if we want to have separate names between upstream/downstream. The upstream/downstream observability-resources-mk config would have different names for the user facing metrics. 

+1.  

> Do we have an upstream name yet?

As far as I remember `kclerk` was the most voted one but I've seen mention of `bf2` (e.g org.bf2, bf2.dev etc) in some other projects. @emmanuelbernard might help us here.

> Otherwise, we can have a generic name for both that's not product specific. Do you have anything in mind?

Other than the `kafka` prefix, I do not have anything in mind.

Maybe @tombentley and @emmanuelbernard have a better ideas or using `rhosak` prefix is just fine overall.

JameelB:
bump for review @pb82 @tombentley @k-wall @StevenTobin @riccardo-forina

JameelB:
> Thank you for writing this @JameelB, it's a big +1 from me. It's been a bit painful to understand which metrics to use to build the dashboard, and the naming convention we have now doesn't help. Any effort in improving the DX has the green light in my book.
> 
> The only bit I'm not sure I like is the `rate5m` part for rate metrics like `|rhosak_incoming_bytes_rate5m`. As a user, I'd probably like to ask for the aggregation through a parameter, but I understand the challenge this would force us to address.

Thanks for the review @riccardo-forina! Allowing the aggregation period to be specified via a parameter can certainly be separated from this adr. But what do you think of not including this in the name? Instead of  `rhosak_incoming_bytes_rate5m`, it'll just be `rhosak_incoming_bytes_rate`. The aggregation period for rate metrics will be documented for now, @tombentley also mentioned this at a previous point.

> how important it is to the customer to know the aggregation period? If so, should it be part of the metric name, or is is sufficient to have it in the documentation?

If we don't include it in the name, it would provide us flexibility to allow users to specify the aggregation period at a later stage when we decide to do that without having to change the name of the metric. For other services that wish to use this pattern, they could also consider this feature in their initial design of their own metrics API.

riccardo-forina:
I think we can keep this for a future ADR if needed, there is nothing inherently wrong in having the `rate5m` in the name right now.

pb82 (approved):

_adr/87/index.adoc
```diff
@@ -1,48 +1,442 @@
...
+
+. Document the new metric
+** This should be assigned to the Documentation/Customer Content Support team.
+** [.underline]#Definition of Done#:
+*** New metric documented in the app-services-guides: https://github.com/redhat-developer/app-services-guides/tree/main/docs/kafka/metrics-monitoring-kafka[Monitoring metrics in OpenShift Streams for Apache Kafka].
+
+
+### Migrating existing metrics to the new naming policy
+KAS Fleet Manager does not support API versioning at the moment so we would need to support both existing and new metrics during the migration process.
+Since the revised metric names will be added as new metrics, they will have no history.
```
tombentley, line 384: So we're providing the new metrics from the same endpoint as the old metrics? That's a much harder migration story for customers, because they have to check each metric individually whether it's going to be removed. 

I know you're defined this versioning problem as a non-goal, but I wonder whether we should think more about how to solve it, because it's intrinsic to providing a clear migration path. Is there a discussion somewhere about why this is difficult/what the issues are?

JameelB: I believe the underlying issue is that kas fleet manager doesn't really support API versioning at the moment. I believe @wtrocki plans to talk about this in the architecture call :)

_adr/87/index.adoc
JameelB: Would there be a preference to use `rhosak_` for downstream or would `kas_` be ok to use in general?

_adr/87/index.adoc
```diff
@@ -1,48 +1,442 @@
...
+|broker
+
+|Topic
+|Metrics associated to Kafka topics
+|None (the topic metrics already have a 'topic' label so we should avoid using this in the metric name, as per the naming policy)
+|===
+
+Examples of sub-component usage in the metric name:
+
+* For *Cluster* metrics, the metric prefix would be `rhosak_`.
```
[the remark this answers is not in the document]
JameelB: yep we can. Is my terminology here correct? It's called a `Kafka cluster` rather than a `Kafka instance`?

_adr/87/index.adoc
JameelB: rhobs team got back to us confirming that there is nothing to worry about the current state of metrics ingestion in Observatorium. We are far off from hitting any limits. Therefore there should be no problem with keeping deprecated metrics for a minimum of 60 days. We could increase this number as well if needed.

JameelB: Is 60 days enough for deprecation period?

emmanuelbernard: So there are two areas to consider:
* a metric name is an API contract and therefore should not be deprecated unless it is really problematic
* a metric content is naturally time bound and we could consider that a deprecation time wound be > a the time out period of a metric entry

JameelB: Thanks @emmanuelbernard!

wtrocki: It depends on what we really productize on the metrics side for users. 
Right now we provide and document to ur customers an Kafka Dashboard dashboard that does provide additional abstraction layer for integrators:

https://grafana.com/grafana/dashboards/15835

We say that when users use that dashboard they would get updates on the configuration automatically thus if something breaks on their side they can do update and review it

That is why I think we should move contract to the client side artifacts (dashboards, grafana example datasets and SDKs. 

Current dashboard and Kafka SDK does that well. 

The main remaining element for our team is missing is the separation of the http metrics. We need new tag in the openAPI file to provide dedicated prometheus http SDK for our end users.

_adr/87/index.adoc
```diff
@@ -1,48 +1,232 @@
...
+
+
+### Migrating existing metrics to the new naming policy
+KAS Fleet Manager does not support API versioning at the moment so we would need to support both existing and new metrics during the migration process.
+Since the revised metric names will be added as new metrics, they will have no history.
+This will need to be taken into account during the migration.
+
+The Kafka console provides dashboards which can display data up to the last 7 days (see the design in https://redhat-developer.github.io/app-services-ui-components/?path=/story/kafka-metrics--all-ready[Storybook]).
+Once 7 days has passed, since the new metrics has been added, the existing and new metrics would have the same history for the UI.
+At this point, we can expose the new metrics in KAS Fleet Manager and mark the existing metrics as deprecated.
```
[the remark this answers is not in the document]
pb82: There will probably be sufficient time between we add those metrics and the UI starts working on this. We won't need to delay adding them to the fleet manager.

_adr/87/index.adoc
JameelB: as per discussion in the architecture call, we will be using `kas_` for both upstream and downstream.

_adr/87/index.adoc
```diff
@@ -1,48 +1,442 @@
...
+However, the Kafka Integrations team would have the knowledge of what metrics to use and how it should be aggregated, if required.
+All discussions around which metric to use and how its aggregated is done via ad-hoc chat conversations, in the JIRA issue or the Github pull request.
+Discussion to reach a formal agreement cross teams for the metric name is not done.
+
+
+### Naming scheme
+The names of the current customer facing metrics are based on the following:
+
+* Raw Kafka metrics: The names of these metrics are not changed, they are taken directly from Kafka metrics. These have the following name format:
+** For example: `kafka_controller_kafkacontroller_offline_partitions_count`
```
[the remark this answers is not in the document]
pb82: @k-wall defining the correct names for the metrics in the Kafka CR instead of renaming them later via rules in the resources repository seems preferable, would you agree?

_adr/87/index.adoc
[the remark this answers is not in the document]
JameelB: sounds good, i'll update this section with that. thanks!

_adr/87/index.adoc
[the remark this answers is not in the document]
pb82: @k-wall do you know what naming is communicated to our internal customers? Do they call it Kafka clusters of Kafka instances?

_adr/87/index.adoc
machi1990: +1 on that as well and thanks for the update @JameelB and for following up on this one.

_adr/87/index.adoc
JameelB: > Discuss and figure out how prometheus "api" would be versioned.

This is listed as a non-goal. See https://github.com/bf2fc6cc711aee1a0c2a/architecture/pull/64#discussion_r887009810

> Figure out where the list of metrics is and likely point to it from the OpenAPI is

Out of scope of this ADR. JIRA created for this task: [MGDSTRM-9106](https://issues.redhat.com/browse/MGDSTRM-9106)

_adr/87/index.adoc
JameelB: All references of the revised metric names has now been updated to use `kas_` prefix.

_adr/87/index.adoc
```diff
@@ -1,48 +1,442 @@
...
+|Topic
+|Metrics associated to Kafka topics
+|None (the topic metrics already have a 'topic' label so we should avoid using this in the metric name, as per the naming policy)
+|===
+
+Examples of sub-component usage in the metric name:
+
+* For *Cluster* metrics, the metric prefix would be `rhosak_`.
+* For *Broker* metrics, the metric prefix would be `rhosak_broker_`.
+* For *Topic* metrics, the metric prefix would be `rhosak_`.
```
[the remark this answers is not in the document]
JameelB: thanks, i've updated this now to be `rhosak_topic`

_adr/87/index.adoc
```diff
@@ -1,48 +1,453 @@
...
+The existing metrics shall be marked as deprecated as per the MAS API deprecation policy.
+The deprecation will be communicated as follows:
+
+* The KAS Fleet Manager OpenAPI specification will document the deprecated metrics and their alternatives in each of the endpoint descriptions.
+* The metrics listed in the https://github.com/redhat-developer/app-services-guides/tree/main/docs/kafka/metrics-monitoring-kafka[Monitoring metrics in OpenShift Streams for Apache Kafka] documentation shall be marked as 'deprecated', pointing the users to use the newly revised metrics.
+* An email will be sent out to mas-devel and mas-all mailing list as per the deprecation policy.
+* Deprecation period shall be 60 days.
+** The metrics API is currently part of the KAS Fleet Manager API lifecycle tag.
+The KAS Fleet Manager API is in beta (no other tags are available at the moment).
+For beta, the deprecation period should be 60 days.
```
JameelB, line 408: Am I right in saying this? We haven't really specified what version our API is in. Should it be stable instead? If that's the case, the deprecation period should probably be 180 days. 

@pb82 @davidffrench

pb82: @JameelB where does the beta tag come from? I'm not seeing it in our OpenAPI file.

JameelB: It was mentioned in the architecture call. See  https://github.com/bf2fc6cc711aee1a0c2a/architecture/pull/64#discussion_r914837142. This isnt in our openapi spec/api path. We never really specified  if our api is in beta/stable version.

JameelB: We could probably include this in the discussion of the api versioning though, so may be out of scope for this. 

For now, since 60 days was agreed on for the deprecation period of the old metrics we could state that here, different deprecation periods for beta/stable version could be something to think about in the future. wdyt?

pb82: Agree, 60 days seems plenty. The UI only allows you to go back 7 days, so this would only affect customers using the federate endpoint.

_adr/87/index.adoc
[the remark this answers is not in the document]
k-wall: Yes, it is possible to do that at source

_adr/87/index.adoc
```diff
@@ -1,48 +1,452 @@
...
+### Kafka Metric Categories
+The customer facing Kafka metrics can be divided into multiple categories.
+Categorizing these metrics into logical groups can help us formulate the metric name and document it.
+
+|===
+|Category | Description | Sub-component (to be specified in the metric name)
+
+|Cluster
+|General Kafka metrics not specific to any other sub-components.
+|cluster
```
k-wall, line 119: I would suggest that we should be using the terminology `instance` rather than `cluster`.  The fact that we are giving each instance its own cluster is really an implementation detail.

_adr/87/index.adoc
```diff
@@ -1,48 +1,452 @@
...
+|===
+
+Examples of sub-component usage in the metric name:
+
+* For *Cluster* metrics, the metric prefix would be `kas_cluster_`.
+* For *Broker* metrics, the metric prefix would be `kas_broker_`.
+* For *Topic* metrics, the metric prefix would be `kas_topic_`.
+
+### Converted Metrics
+At minimum, each metric should have an `instance_name` label that identifies the Kafka instance if possible.
```
k-wall, line 137: +1

k-wall: Should we also expose the instance id?

_adr/87/index.adoc
```diff
@@ -1,48 +1,452 @@
...
+|Gauge
+|Not available
+|instance_name
+|Maximum number of partitions per instance
+
+|kafka_namespace:haproxy_server_bytes_in_total:rate5m
+|kas_cluster_incoming_bytes_rate5m
+|Gauge
+|Not available
+|Not available (no label available to get the instance_name from)
```
k-wall, line 163: The metric needs the label to be useful to the customer.  It is the implementation's job to find the right way to deliver it.

_adr/87/index.adoc
```diff
@@ -1,48 +1,452 @@
...
+|Old Metric Name | Revised Metric Name | Type | Old Labels | Revised Labels | Description
+
+|kubelet_volume_stats_available_bytes
+|kas_broker_volume_available_bytes
+|Gauge
+|persistentvolumeclaim
+|volume (source label: persistentvolumeclaim)
+|Number of available disk space, in bytes, per broker
+
+|kubelet_volume_stats_used_bytes
```
k-wall, line 228: duplicates kas_broker_storage_used_bytes

_adr/87/index.adoc
```diff
@@ -1,48 +1,452 @@
...
+|instance_name
+|Number of desired brokers per instance
+
+|===
+
+#### Broker Metrics
+|===
+|Old Metric Name | Revised Metric Name | Type | Old Labels | Revised Labels | Description
+
+|kubelet_volume_stats_available_bytes
```
k-wall, line 221: do we actually need this metric?.  we have kas_broker_storage_limit_bytes and kas_broker_storage_used_bytes - that is sufficient.

_adr/87/index.adoc
```diff
@@ -1,48 +1,452 @@
...
+
+|kafka_broker_quota_totalstorageusedbytes
+|kas_broker_storage_used_bytes
+|Gauge
+|broker_id, statefulset_kubernetes_io_pod_name, strimzi_io_cluster
+|broker_id, instance_name (source label: strimzi_io_cluster)
+|Number of storage used by partitions, in bytes, per broker
+
+|kafka_broker_quota_hardlimitbytes
+|kas_broker_storage_hard_limit_bytes
```
k-wall, line 243: we don't distinguish between hard and soft limit anymore.  suggest simply: `kas_broker_storage_limit_bytes`

_adr/87/index.adoc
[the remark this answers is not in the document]
k-wall: @pb82 we use the term `instance` on the RHOSAK documentation https://console.redhat.com/beta/application-services/learning-resources.  The dataplane as refers to `'instance`.   The fact that an instance is currently provided by a single kafka cluster is really an implementational detail.

_adr/87/index.adoc
```diff
@@ -1,48 +1,452 @@
...
+|Maximum message size, in bytes, supported by the instance.
+
+|kafka_instance_partition_limit
+|kas_cluster_partition_limit
+|Gauge
+|Not available
+|instance_name
+|Maximum number of partitions per instance
+
+|kafka_namespace:haproxy_server_bytes_in_total:rate5m
```
k-wall, line 159: FYI - there is an defect affecting the implementation of this metric at the moment, tha means the values produce by it and `kafka_namespace:haproxy_server_bytes_out_total:rate5m` can be incorrect. https://issues.redhat.com/browse/MGDSTRM-9131

_adr/87/index.adoc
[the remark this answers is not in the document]
JameelB: Thanks for confirmation @k-wall, I'll update this category to be `Instance` and have the prefix `rhosak_instance_`

_adr/87/index.adoc
JameelB: When retrieving metrics from the API, you'd need to specify the Kafka ID. In this case, the instance id should already be known. However, one case where I think this would be useful for is when these metrics are integrated to a metric platform. It would help identify which kafka instance the metric belongs to if a user has federated multiple Kafka metrics to one metric platform. 

@k-wall  Is it possible to expose the instance id as part of the raw kafka metrics?

_adr/87/index.adoc
JameelB: Are these two metrics the same then `kafka_broker_quota_hardlimitbytes` and `kafka_broker_quota_softlimitbytes`?

_adr/87/index.adoc
JameelB: will remove, thanks!

_adr/87/index.adoc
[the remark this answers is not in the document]
JameelB: Would an epic brief be the right place to discuss/have proposed solutions for how we may do this? @pb82 @k-wall @tombentley

_adr/87/index.adoc
JameelB: I'm not too sure but it could be handy to not do any additional calculation to get this value? @riccardo-forina are you using this metric in the UI?

riccardo-forina: No, but we point to that metric [in the guides](https://github.com/redhat-developer/app-services-guides/blob/2bfe9b92ab23f2002063581adae9b92bd65eede9/docs/kafka/metrics-monitoring-kafka/README.adoc#L128)

JameelB: thanks for confirming. I think we're ok to leave this in for now since it's already there, unless there's a strong need to remove it.

riccardo-forina: Sounds good!

JameelB:
@k-wall I've addressed your comments, please have a look again when you get a chance. 

@tombentley @StevenTobin @wtrocki Can you take a look at this when you get a chance too? Would like to get your approval here :)

wtrocki:
LGTM

_adr/87/index.adoc
```diff
@@ -1,48 +1,451 @@
...
-// Bulleted list of outcomes that this ADR is not trying to achieve.
+- Metrics API versioning
++
+The metrics API is currently part of the KAS Fleet Manager API.
+We need to consider KAS Fleet Manager API as a whole in order to have consistent versioning strategy for all of our endpoints, including customer facing metrics and the metrics API.
++
+
++
+This also includes different deprecation periods with different lifecycle tags and the idea of separating the metrics API from the rest of the KAS Fleet Manager API to have isolated versioning.
+This is out of scope of this ADR and should have its own separate discussion, epic or ADR.
```
tombentley, line 63: I'm fine that this is out of scope, but I do think this is a conversation worth having, because it could be that other managed services fall into the same trap. And perhaps there's other functionality than metrics which ought not to be caught up into the same scope of versioning as the rest of the API.

_adr/87/index.adoc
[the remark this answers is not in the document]
tombentley: Yes, Epic Brief would be the right place.

tombentley (approved):
Thanks for this @JameelB LGTM.

_adr/87/index.adoc
JameelB: Thanks for reviewing @tombentley! 

Agreed. I've added this as a topic to our team's tech call to start initial discussions around this. cc @pb82

StevenTobin (approved):

_adr/87/index.adoc
k-wall: Yes -  we've got several options.  One option is to take the instance id out of the kafka namespace (and removed the kafka- portion).  There's probably a better option to put the id on the pod resources as a label, then rely on the fact that kube labels belome prometheus labels.

_adr/87/index.adoc
k-wall: Yes.   From point of view of the public API I would just expose 'limit bytes'.

_adr/87/index.adoc
k-wall: That defect is fixed in production.

_adr/87/index.adoc
pb82: We use options #1 a lot with the billing metrics: `label_replace` can take the id from the namespace and put it into a new label.

_adr/87/index.adoc
```diff
@@ -1,48 +1,451 @@
...
+
+|kafka_instance_partition_limit
+|kas_instance_partition_limit
+|Gauge
+|Not available
+|instance_name, instance_id
+|Maximum number of partitions per instance
+
+|kafka_namespace:haproxy_server_bytes_in_total:rate5m
+|kas_instance_incoming_bytes_rate5m
```
k-wall, line 161: I questioning the final underscore separator in metrics that convey rates.  By convention Prometheus uses `:rate_xt_`

in the `_rate5m` suffix. By convention the las    Prometheus's convention is colon.   

kas_instance_outgoing_bytes_total
kas_instance_outgoing_bytes:rate5m

_adr/87/index.adoc
```diff
@@ -1,48 +1,451 @@
...
+* For *Broker* metrics, the metric prefix would be `kas_broker_`.
+* For *Topic* metrics, the metric prefix would be `kas_topic_`.
+
+### Converted Metrics
+At minimum, each metric should have a label that identifies the Kafka instance if possible.
+This would be useful if a user has multiple Kafka instances.
+If they want to integrate all of their Kafka metrics into one monitoring system, the label(s) can act as an identifier.
+For Kafka metrics, we can use `instance_name` and `instance_id` to identify the Kafka instance the metric belongs to.
+
+#### Cluster Metrics
```
k-wall, line 142: I wouldn't include the Converted Metrics as part of the ADR.  I would describe the naming conventions and back those up with a few good examples (examples from the managed kafka domain is fine).  I'd move the Converted Metrics section to the Epic Brief.  Others might disagree.

_adr/87/index.adoc
JameelB: In Prometheus convention, the `:` are reserved for user defined recording rules. It may be often used for rates because they would usually be used for these. Do we want to mix this in our label name? We will have the following types then:

* kas_\<metric\>_\<unit: e.g. bytes|seconds\>
* kas_\<metric\>_\<type/aggregate: e.g. total|info|bucket\>
* kas_\<metric\>:ratent

The suffix may not always be a unit either, it could be a metric type or aggregate i.e. `total` or `info`. Should this rule apply to those as well?

_adr/87/index.adoc
JameelB: I've no problem moving these to the epic brief. wdyt @tombentley @pb82?

_adr/87/index.adoc
JameelB: Thanks for the update on this @k-wall

_adr/87/index.adoc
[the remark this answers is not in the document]
JameelB: Epic brief has been created for this epic. It's linked in the description of the epic issue in JIRA: [MGDSTRM-7080](https://issues.redhat.com/browse/MGDSTRM-7080)

_adr/87/index.adoc
JameelB: I've added this topic to the epic brief that's linked in the description of this epic in JIRA [MGDSTRM-7080](https://issues.redhat.com/browse/MGDSTRM-7080)

_adr/87/index.adoc
JameelB: Thanks for confirming @k-wall, I've included `kafka_broker_quota_softlimitbytes` in the list of metrics to be removed.

_adr/87/index.adoc
JameelB: Epic brief has been created, which is linked to this issue in JIRA: https://issues.redhat.com/browse/MGDSTRM-7080. 

I've moved the converted metrics into that and removed it from the ADR.

pb82:
@k-wall are you Ok with approving and merging this ADR? As discussed, implementation details will be in the form of an Epic Brief.

k-wall (approved):
lgtm

_adr/87/index.adoc
k-wall: Thanks for the feedback. On reflection, let's go with your idea.

tombentley:
@machi1990 and @emmanuelbernard are the remaining reviewers who haven't LGTM'd. Any further feedback, or can we now merge this?

machi1990 (approved):

JameelB:
bumping @tombentley @emmanuelbernard :)

tombentley:
I've LGTM'd. Emmanuel _might_ be able to look at it towards the end of this week and it might be good if he could have another look, but if you're blocked on this then I'm happy to merge it. Let me know!

JameelB:
> I've LGTM'd. Emmanuel _might_ be able to look at it towards the end of this week and it might be good if he could have another look, but if you're blocked on this then I'm happy to merge it. Let me know!

Not blocked on this, can wait for Emmanuel to take a look :) Thanks @tombentley!

_ap/16/index.adoc
```diff
@@ -13,56 +13,114 @@ tags:
...
+The metric name should adhere to the format below:
 
-## Participants
+*<service-name>_<optional:sub-component>_<descriptor>_<optional:base-unit>_<optional:type|aggregate>*
 
-// What components and/or teams are directly involved
-//
-// For this metrics example this would include the customer and the control plane via which
-// the metrics are exposed
+* *service-name*: The metric name should be prefixed with the service name (e.g. `kas`)
```
emmanuelbernard, line 45: This does not seem to be the service name but rather the service prefix (e.g. `kas`). Maybe add something a bit more clear in name and description

_adr/87/index.adoc
```diff
@@ -1,48 +1,227 @@
...
+* *Consistency*: These two types of metrics follow different naming schemes, resulting in inconsistent metric names.
+* *Repetition*: The raw Kafka metrics contain a lot of repetition i.e. `kafka_controller_kafkacontroller...`.
+* *Exposure of implementation details*: The name of the recording rules metrics exposes implementation details i.e. `kafka_namespace:haproxy_server_bytes_in_total:rate5m`, end users do not need to know about namespaces and OpenShift.
+
+Since we don't have a formal process and naming policy for adding customer facing metrics, we often reach a situation where the resulting metric names are confusing and misleading (i.e. https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/issues/688[misleading metric name for kafka topic partitions count]).
+
 
 ## Goals
-// Bulleted list of outcomes that this ADR, if accepted, should help achieve
+- Define a naming policy for existing and new customer facing metrics.
```
emmanuelbernard, line 49: The doc seems to express that KAS fleet manager is proxying Observatorium meaning it express the implementation choices.

However this does not seem to be part of the ADR goal. Should it be split as a new aDR or simply added to this ADR?

cc @tombentley for advice here.

_adr/87/index.adoc
```diff
@@ -1,48 +1,227 @@
...
+The description of this issue should have sufficient information on what this new metric should provide in order to identify the metric to be used and how it should be documented in later stages.
+
+The issue created must have the following sub-tasks and they must be completed in order:
+
+. Identify the metric and labels to be used and how it will be aggregated if required.
+** This should be assigned to the team that has the knowledge of the requested metric.
+** [.underline]#Definition of Done#:
+*** Metric to be used is identified.
+*** (Optional) Aggregation of the metric determined.
+*** (Optional) Category/Sub-component which this new metric belongs to is identified.
```
emmanuelbernard, line 152: This is the only place where we mention Category/Sub-component, elsewhere in the two documents, it is only mentioning Sub-component. I think we should keep it homogeneous.

This leads to my more general question. Is sub-component a correct model? I understand we don't mention the technology names like Strimzi, still the sub-components are possible implementation details of the service. But by exposing them, we are making them aprt of the customer contract and thus be documented. So instance, brokers and topics are concepts we want exposed in the API?

If yes, that's good but in the AP and ADRs, I would still add a caution block explaining ththat we do not want to expose things we consider implementation details that we would reserve the right to change later as names for sub-components.

emmanuelbernard (changes requested):
I've added my review with minor comments and one philosophical question. When processed, it's lgtm for me.

_adr/87/index.adoc
tombentley: Good point. I think topics and brokers are foundational concepts in a Kafka service, so it's impossible to avoid the manifesting in this API. But we should be very careful about things which are not intrinsic to the service being provided. For example, metrics from other components. Those will often be _about_ the foundational concepts (like topics and brokers) but their _origin_ (say, canary or kafka exporter) is an implementation detail which shouldn't be exposed in the API.

_adr/87/index.adoc
tombentley: I don't think a separate ADR would necessarily be needed (ultimately I see this as being about how we expose metrics as an API, naming is one part of that and proxying is another). However if we kept it in this ADR we would need to adjust the title, goals etc to make the scope clear.

emmanuelbernard: I'm cool with a title adjustment.

JameelB: Does this title sound reasonable for this adr: `Exposing metrics via the KAS Fleet Manager API`?

_adr/87/index.adoc
JameelB: > This is the only place where we mention Category/Sub-component, elsewhere in the two documents, it is only mentioning Sub-component. I think we should keep it homogeneous.

Would category make more sense in terms of terminology than a sub-component? Category is what we've been using to group these metrics in our documentation. Topics/brokers would probably makes sense as a 'category' rather than a 'sub-component' too. 

> If yes, that's good but in the AP and ADRs, I would still add a caution block explaining ththat we do not want to expose things we consider implementation details that we would reserve the right to change later as names for sub-components.

> Those will often be about the foundational concepts (like topics and brokers) but their origin (say, canary or kafka exporter) is an implementation detail which shouldn't be exposed in the API.

Good point, i'll add a caution block in these sections to make sure we do not expose implementation details.

emmanuelbernard: Category intuitively works for me but I haven't played much with the concepts.

_adr/87/index.adoc
JameelB: Title changed, thanks!

_ap/16/index.adoc
JameelB: Updated to use service prefix instead.

emmanuelbernard (approved):
I gave the ok after change reflect the last round comment conversations. So optimistic approve here.

JameelB:
Thanks for the reviews! No other changes on my side. Snce all reviewers have LGTM'ed, please go ahead and merge when you get a chance @tombentley @emmanuelbernard

PR #67, opened 2022-06-01
Add the contents for ADR 86 - Canary authentication with sso.redhat.com

Rajagopalan-Ranganathan:
JIRA: https://issues.redhat.com/browse/MGDSTRM-8157

Detail the existing authentication of Strimzi Canary in our RHOSAK
env with MAS SSO and the proposed changes for the authentication flow
with sso.redhat.com.


Rajagopalan-Ranganathan:
@akoserwal @MikeEdgar could you take a look when you find time. thanks!

akoserwal:
_adr/86/adr-86-proposed-authentication-flow.png validate orgId won't the case for canary service accounts. In general org id check is present in the custom claim check. The arrow for the protected resource should be pointing the broker?

akoserwal:
_adr/86/adr-86-proposed-authentication-flow.png validate orgId won't the case for canary service accounts. In general org id check is present in the custom claim check. The arrow for the protected resource should be pointing the broker?

Rajagopalan-Ranganathan:
> _adr/86/adr-86-proposed-authentication-flow.png validate orgId won't the case for canary service accounts. In general org id check is present in the custom claim check. The arrow for the protected resource should be pointing the broker?

@MikeEdgar can you confirm if the auth flow only checks for the existence of an orgID or does it validate against the orgID of the creator as well?

MikeEdgar:
> @MikeEdgar can you confirm if the auth flow only checks for the existence of an orgID or does it validate against the orgID of the creator as well?

@Rajagopalan-Ranganathan it validates (currently) that the orgId matches that of the creator.
```
customClaimCheck: '@.rh-org-id == ''testorgid''|| @.org_id == ''testorgid'''
```

_adr/86/index.adoc
```diff
@@ -13,21 +13,88 @@ tags:
...
+
+This would mean that all the Canary service accounts for all customers/users will be created/associated with this "Special Management Organization" when compared to the existing architecture where they were scoped to the user's organization. This requires some additional steps in authentical flow described in the current architecture section.
+
+- KFM when building the "CustomClaimCheck" includes the  Canary service account client ID "clientID" field as well
+
+```
+   - authentication:
+      customClaimCheck: '@.rh-org-id == ''testorgid''|| @.org_id == ''testorgid'' || @.clientId == ''canar-client-Id'''
```
MikeEdgar, line 62:
```suggestion
      customClaimCheck: '@.rh-org-id == ''testorgid''|| @.org_id == ''testorgid'' || @.clientId == ''canary-client-Id'''
```

_adr/86/index.adoc
```diff
@@ -13,21 +13,88 @@ tags:
...
+      password: **********
+      principal: canary-client-Id    
+```
+
+
+- Token Endpoint URI would now point to sso.redhat.com
+- The existing authentication flow up until receiving a valid token from the service mentioned in Token end point URI continues as it is.
+- During the Token introspection and performing the custom claim check validation, the following will be done:
+  - Verify that the "Canary service account Client" belongs to the "Special Management Org"
+  - Verify that the "ClientID" in the custom claim check matches the expected Canary client ID
```
MikeEdgar, line 77: This doesn't match the proposed claim check above. I think it should continue verifying that the token belongs to the org that created the cluster -OR- verify that the ClientID is the canary client

_adr/86/index.adoc
Rajagopalan-Ranganathan: fixed

_adr/86/index.adoc
Rajagopalan-Ranganathan: good catch Mike. I have updated the text and diagram to reflect the same. For canary the authn will be match the clientID since the orgID claims cannot be matched.

MikeEdgar: @Rajagopalan-Ranganathan - regarding the custom claim check, should this doc still mention that the orgID check is present (for tokens other than the canary's)? I just want to confirm the intent. Right now in the diagram and in this list it looks like it will be removed.

Rajagopalan-Ranganathan: OrgID will be present for other checks, this is about the canary authn part alone. if required i can amend the text accordingly.

MikeEdgar: Thanks for clarifying!

MikeEdgar (approved):

_adr/86/index.adoc
```diff
@@ -13,21 +13,87 @@ tags:
...
 
 ## Current architecture
 
+Currently, Kas Fleet Manager (KFM) creates a Canary service account client in MAS SSO upon a request to create and provision a Kafka cluster. These Canary service account clients are associated with the same organization as the user making the request to create a Kafka cluster. This information is contained in the `customClaimCheck` field in the Kafka Custom Resource (CR)
+
+```
+   - authentication:
+      customClaimCheck: '@.rh-org-id == ''testorgid''|| @.org_id == ''testorgid'''
+      tokenEndpointUri: >-
+            https://tokenendpointauri.com/token
```
tombentley, line 37: Just FYI [RFC-2606](https://www.rfc-editor.org/rfc/rfc2606.html) provides `example.com` for this kind of purpose.

_adr/86/index.adoc
```diff
@@ -13,21 +13,87 @@ tags:
...
+
+This would mean that all the Canary service accounts for all customers/users will be created/associated with this "Special Management Organization" when compared to the existing architecture where they were scoped to the user's organization. This requires some additional steps in authentical flow described in the current architecture section.
+
+- KFM when building the "CustomClaimCheck" includes the  Canary service account client ID "clientID" field as well
+
+```
+   - authentication:
+      customClaimCheck: '@.rh-org-id == ''testorgid''|| @.org_id == ''testorgid'' || @.clientId == ''canary-client-Id'''
```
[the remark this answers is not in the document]
akoserwal: it will fail for external clients/users as they will only satisfy the org id check. Canary accounts is going to have different org id.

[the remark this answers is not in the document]
Rajagopalan-Ranganathan: That is correct.. All internal service accounts will not be a part of Customers Org ID. Resolving this.

_adr/86/index.adoc
```diff
@@ -13,21 +13,87 @@ tags:
...
+
+Upon a successful authentication, MAS SSO provides a valid token back to the Kafka broker. Kafka brokers as a part of its authentication flow, upon a successful response, it introspects the token received and performs a custom claims check. It verifies that the "Canary Service account Client" belongs to the same organization as the creator or owner of the Kafka instance. Kafka brokers deem the authentication to be successful after this.
+
+image::adr-86-current-authentication-flow.png[Current authenticaion flow]
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: Agreed tom. We can work on that on a separate ADR of sso.redhat.com auth services for MAS/RHOSAK and cover the overall role of sso.redhat.com in MAS/RHOSAK.. WDYT?

_adr/86/index.adoc
Rajagopalan-Ranganathan: thanks Tom.. That is useful information.

Rajagopalan-Ranganathan:
@akoserwal @MikeEdgar @tombentley let me know if this is in good state to be merged. Mike has given LGTM already.

akoserwal:
lgtm

_adr/86/index.adoc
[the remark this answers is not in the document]
tombentley: So I might have got me example wrong, but surely the `@.clientId == ''canary-client-Id'''` should apply only when the org id _is also_ this special RH internal org for the service account? What about an account with `clientId == canary-client-id`, but in a different org?

Rajagopalan-Ranganathan:
> Apart from the question just raised, I think this is written a little too much in the style "Red Hat this and Red Hat that". It would be good to rephrase it to be a bit more generic (e.g. talk about "Service Provider"). I realise this is trickier than most other ADRs because the SSO integration is more Red Hat specific that other things. Wdyt?

I had a look, and most of the red hat comes from "sso.redaht.com", unfortunately this is very specific to red hat sso and cannot be generalized for another sso system even within redhat at the moment. I would prefer this one to be specific about RH SSO as the principles in general discussed are heavily tied to the same.

tombentley:
@Rajagopalan-Ranganathan that makes sense; thanks for taking the time to look at it/

_adr/86/index.adoc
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: Since all Canaries will be created in the same Org (Special management Org) that seemed a bit redundant in our use case. All internal service accounts are created in "Special management ORg"

tombentley (approved):

PR #70, opened 2022-07-04
Fix #40 Add a glossary page and a mechanism to link to glossary items

### LINKED ISSUE #40: Define personas
ADRs often needs to refer to groups/teams or people with a particular function (e.g.  "customer support team", "support engineer" etc). We shouldn't use some private jargon for these things, so they need generic terms and to be explained somewhere.

k-wall:
… from ADRs etc.

Add some persons as glossary items.

### DIFF
--- a/_includes/glossary_links.adoc
+++ b/_includes/glossary_links.adoc
@@ -1,0 +1,6 @@
+// glossary_links.adoc
+// The attributes defined provide links to anchors on the glossary page.  Please also remember to add the
+// corresponding definition to glossary.adoc.
+:glossary-businessstakeholder-: link:/glossary#glossary-businessstakeholder[Business Stakeholder]
+:glossary-customersupport: link:/glossary#glossary-customersupport[Customer Support]
+:glossary-sre: link:/glossary#glossary-sre[SRE]
--- a/_includes/glossary_links.adoc
+++ b/_includes/glossary_links.adoc
@@ -1,6 +1,6 @@
 // glossary_links.adoc
 // The attributes defined provide links to anchors on the glossary page.  Please also remember to add the
 // corresponding definition to glossary.adoc.
-:glossary-businessstakeholder-: link:/glossary#glossary-businessstakeholder[Business Stakeholder]
+:glossary-businessstakeholder: link:/glossary#glossary-businessstakeholder[Business Stakeholder]
 :glossary-customersupport: link:/glossary#glossary-customersupport[Customer Support]
 :glossary-sre: link:/glossary#glossary-sre[SRE]
--- a/_adr/0/index.adoc
+++ b/_adr/0/index.adoc
@@ -10,10 +10,12 @@ applies_padrs: # What PADRs does this ADR apply?
 applies_patterns: # What APs does this ADR apply?
 ---
 
+include::/_includes/glossary_links.adoc[]
+
 // Top style tips:
 // * Use one sentence per line
 // * No unexpanded acronyms
-// * No undefined jargon
+// * No undefined jargon - use references to the glossary as appropiate e.g. {glossary-sre}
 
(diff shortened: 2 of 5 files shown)


_includes/glossary_links.adoc
```diff
@@ -0,0 +1,6 @@
+// glossary_links.adoc
+// The attributes defined provide links to anchors on the glossary page.  Please also remember to add the
+// corresponding definition to glossary.adoc.
+:glossary-businessstakeholder-: link:/glossary#glossary-businessstakeholder[Business Stakeholder]
```
grdryn, line 4:
```suggestion
:glossary-businessstakeholder: link:/glossary#glossary-businessstakeholder[Business Stakeholder]
```

_includes/glossary_links.adoc
```diff
@@ -0,0 +1,6 @@
...
+:glossary-customersupport: link:/glossary#glossary-customersupport[Customer Support]
+:glossary-sre: link:/glossary#glossary-sre[SRE]
```
grdryn, line 6:
```suggestion
:glossary-sre: pass:[<abbr title="Site Reliability Engineer">link:/glossary#glossary-sre[SRE]</abbr>]
```

^Just a thought: this would give hover text to the abbreviation. It's still linked to the glossary for people to click-through for more info.

grdryn (commented):
Would it be useful for this PR (and future PRs that introduce new jargon), to actually update existing pages to use the glossary references?

_includes/glossary_links.adoc
k-wall: I did wonder about using that technique so that you could hover over any glossary term and the hover text would show the expansion.    To so that I would need a second attribute containing the glossary item's description which would be referenced on pages/glossary.adoc and to give the hover text.  I decided that was over-kill.

I think you are suggesting a special case for SRE as it is an abbreviation.

k-wall:
> Would it be useful for this PR (and future PRs that introduce new jargon), to actually update existing pages to use the glossary references?

This PR will just establish the mechanism.
I don't think Accepted ADR etc should be updated.  It would be appropriate for in-flight and new ADRs (etc) adopt the new practice.

emmanuelbernard:
One of the reasons for the glossary is that the Red Hat or any company internal terminology could deviate from the common industry standard accepted terms. And the mechanism would allow to "override" such terms to limit the friction when read within a context. In that context it was not a glossary so much as a term overriding model.

Can you explain how this would be done withe the model you chose? I'm looking for the impact / overhead of the term changes. Maybe a document explaining how could be part of the PR so it can be read.

k-wall:
@emmanuelbernard - The PR is proposing a common place where we can define terms, such as the name of actors, and provide a mechanism so that ADRs etc can reference those terms.  This will help the reader understand what the term means by linking to its definition.  I am hoping this will aid clarity.

The writer of the ADR would use references such as ` {glossary-sre}`.  I included that hint in the ADR template as comments.  When ADRs are reviewed, the reviewers would need to make sure the submission is jargon free, the glossary used/expanded if appropriate.

We'd disallow updates to existing glossary entries to prevent the meaning of existing referees being changed retrospectively.  We could have a mechanism to deprecate terms if need be.  I think we'd strive to keep the number of terms in the glossary small.

emmanuelbernard:
I hear you @k-wall but how does it address the problem of allowing the _name_ itself be replaced by a different name when in an organization like Red Hat. For example QA being the public name and QE being the Red Hat preferred or more specific alternative.

k-wall:
@tombentley this was discussed in 2022-08-29 arch call, @emmanuelbernard indicated he was happy to proceed as it stands.  If you are happy with the approach, can this be merged?

tombentley:
Thanks @k-wall!

PR #74, opened 2022-08-30
adr-88: Authorization for fleet-manager admin APIs

pawelpaszki:
## What
adr-88: Authorization for fleet-manager admin APIs


emmanuelbernard:
What's the status of this PR, to be reviewed?

akoserwal:
@emmanuelbernard draft stage, it require some work

pawelpaszki:
@akoserwal - I have updated the ADR. would you mind taking a look?

_adr/88/index.adoc
```diff
@@ -1,48 +1,64 @@
...
-// * Use one sentence per line
-// * No unexpanded acronyms
-// * No undefined jargon
-
-// No need for a title heading, it's added by the template
+:rover-groups: https://rover.redhat.com/groups
 
 ## Context and Problem Statement
-// What is the background against which this decision is being taken?
+The purpose of this ADR is to define a standardized approach for securing admin API in a fleet manager
```
akoserwal, line 16: please can you add more context about the admin API & roles?

For example- (reword to make them generic)
To list and review information about Kafka instances, the account must have the kas-fleet-manager-admin-read role.

To list Kafka instances or view and update a Kafka instance, the account needs to have the kas-fleet-manager-admin-write role assigned to it.

To list Kafka instances, or view, update and delete a Kafka instance, the account needs to have the kas-fleet-manager-admin-full role assigned to it

pawelpaszki: thanks @akoserwal - I have added the explanation about these roles and their permissions

_adr/88/index.adoc
```diff
@@ -1,48 +1,70 @@
...
 
 ## Goals
-// Bulleted list of outcomes that this ADR, if accepted, should help achieve
+* Provide standardized way to secure admin API endpoints for fleet managers
 
 ## Non-goals
-// Bulleted list of outcomes that this ADR is not trying to achieve.
 
 ## Current situation
```
akoserwal, line 23: Please can you add about the `Current situation.

pawelpaszki: but for generic ADR there isn't really current situation, am I right?

akoserwal: yes, I agree.

pawelpaszki: so anything else to be added here or in this ADR? Can I remove the `draft` from the PR description?

akoserwal: yes, looks good to me. You can remove the draft remark

akoserwal:
@emmanuelbernard @tombentley 
Please review whenever you get a chance.

_adr/88/index.adoc
```diff
@@ -1,48 +1,70 @@
...
-// Where are we now?
 
 ## Proposal
-// What is the decision being proposed
+
+image::adr-88-admin-api-architecture-diagram.png[architecture diagram]
+
+A Keycloak realm should be used to authorize access to admin API endpoint(s).
```
emmanuelbernard, line 29: in this document s/should be/is/g It's easier to read and project the future into the now :)

_adr/88/index.adoc
```diff
@@ -1,48 +1,70 @@
...
 ## Current situation
-// Where are we now?
 
 ## Proposal
-// What is the decision being proposed
+
+image::adr-88-admin-api-architecture-diagram.png[architecture diagram]
+
+A Keycloak realm should be used to authorize access to admin API endpoint(s). 
+Such realm should contain roles that allow access to endpoints to the privileged API users.
```
emmanuelbernard, line 30: s/should/does/

_adr/88/index.adoc
```diff
@@ -1,48 +1,70 @@
...
+Recommended approach would be to have set of roles per environment and name them in the way that indicates the permissions they grant as follows:
```
emmanuelbernard, line 31: to have set of roles?

_adr/88/index.adoc
```diff
@@ -1,48 +1,70 @@
...
+
+* <unique-fleet-manager-prefix>-admin-read-stage
+* <unique-fleet-manager-prefix>-admin-write-stage
+* <unique-fleet-manager-prefix>-admin-full-stage
+
+* <unique-fleet-manager-prefix>-admin-read-prod
+* <unique-fleet-manager-prefix>-admin-write-prod
+* <unique-fleet-manager-prefix>-admin-full-prod
+
+These roles map to the following permissions for the environment included as the role postfix (e.g. `prod`):
```
emmanuelbernard, line 41: add an example of unique fleet managher prefix for readablity. I assume `kas` is an example.

_adr/88/index.adoc
```diff
@@ -1,48 +1,70 @@
...
+
+Once Keycloak instance is set up, a privileged user uses their credentials to get a JWT token from Keycloak.
+The JWT token will be sent to a admin API in a HTTP request.
+Admin API authorization middleware verifies token signature with Public JWKS Certificate and roles included in the JWT token are checked against roles specified in the fleet manager.
+Matching role for given admin endpoint will grant access to the action associated with that endpoint.
+Failure to match the required role will result in restricting access to the endpoint.
+
+image::adr-88-authentication-flow.png[architecture diagram]
+
+## Suggested implementation for Red Hat Internal Keycloak
```
emmanuelbernard, line 55: hum @tombentley , how can we deal with such situation? (internal RHT sauce descriptions? The idea was to use the same model as SOP layering. But i can see it might not be ready.

1. We accept the imperfection
2. we use the old ADR folder model as a addendum `ADR-88 Authorization for fleet-manager admin APIs -- Red Hat Addendum` ina  doc and point to this one + add the red hat specific paragraph below?
3. other

_adr/88/index.adoc
```diff
@@ -1,48 +1,70 @@
...
+
+Instead of using dedicated Keycloak instance/ realm, roles allowing to access admin API endpoints are represented by membership in a {rover-groups}[Rover group(s)]
 
 ### Threat model
-// Provide a link to the relevant threat model. 
-// You must either update an existing threat model(s) to cover the changes made by this ADR, or add a new threat model.
+For both, general and Red Hat specific approaches, there will be very limited amount of privileged users.
```
emmanuelbernard, line 60: you could add that it relies on the existing security model we use for general kafka instance access, so not a new paradygm? I'm assuming what I say is true.

pawelpaszki: well. this is a generic ADR, so I was hesitant to add anything here. would you agree?

_adr/88/index.adoc
pawelpaszki: added kfm example

_adr/88/index.adoc
pawelpaszki: changed

_adr/88/index.adoc
pawelpaszki: changed

_adr/88/index.adoc
pawelpaszki: I have slightly reworded this section. @emmanuelbernard - please let me know if that's clear enough

_adr/88/index.adoc
```diff
@@ -1,48 +1,67 @@
...
 applies_patterns: # What APs does this ADR apply?
 ---
 
-// Top style tips:
-// * Use one sentence per line
-// * No unexpanded acronyms
-// * No undefined jargon
-
-// No need for a title heading, it's added by the template
+:rover-groups: https://rover.redhat.com/groups
```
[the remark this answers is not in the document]
pawelpaszki: I have removed it, but there is still an outstanding comment about this section: `## Suggested implementation for Red Hat Internal Keycloak` which is related to the rover groups. Let me know if that section needs a change as well

_adr/88/index.adoc
emmanuelbernard: I understand that you see the threat model as attached to how red hat deploys the stack, correct?

That is a good point and is fair but some parts are inherently part of the architecture of the upstream. I think what I described is in this category (unless I’m over my head which could very well be).

pawelpaszki: I have rewritten it a bit. does it make sense now?

emmanuelbernard: yes thank you

_adr/88/index.adoc
tombentley: Better, I think, to say what you've said here: "This is an unsolved problem in the general case, so there is no current situation".

_adr/88/index.adoc
```diff
@@ -1,48 +1,71 @@
...
+
+* <unique-fleet-manager-prefix>-admin-read-prod
+* <unique-fleet-manager-prefix>-admin-write-prod
+* <unique-fleet-manager-prefix>-admin-full-prod
+
+These roles map to the following permissions for the environment included as the role postfix (e.g. `prod`):
+
+* \*-read-* - list permissions
+* \*-write-* - list/ update permissions
+* \*-full-* - full permissions (list/ update/ delete)
```
tombentley, line 43: This is difficult to parse. The first is a pattern that matches permissions described above, but I don't really understand what's after the `-`. Perhaps a table would help?

_adr/88/index.adoc
```diff
@@ -1,48 +1,71 @@
...
+
+Example full role for kas-fleet-manager in production environment is: `kas-fleet-manager-admin-full-prod`.
+
+Once Keycloak instance is set up, a privileged user uses their credentials to get a JWT token from Keycloak.
+The JWT token is sent to a admin API in a HTTP request.
+Admin API authorization middleware verifies token signature with Public JWKS Certificate and roles included in the JWT token are checked against roles specified in the fleet manager.
+Matching role for given admin endpoint grants access to the action associated with that endpoint.
```
tombentley, line 50: Can we cross-reference with the sequence numbers in the diagram?

_adr/88/index.adoc
```diff
@@ -1,48 +1,71 @@
...
+Failure to match the required role results in restricting access to the endpoint.
```
tombentley, line 51: What does `restricting assess` mean exactly? HTTP 401?

_adr/88/index.adoc
tombentley: It's a great question. In theory 1 should be fine, because `security != obscurity`. But I think 2 would be pragmatic all the same.

_adr/88/index.adoc
```diff
@@ -1,48 +1,71 @@
...
+## Suggested implementation for Red Hat Internal Keycloak
+
+Instead of using dedicated Keycloak instance/ realm, roles allowing to access admin API endpoints are represented by membership in a Red Hat Rover Group(s).
 
 ### Threat model
-// Provide a link to the relevant threat model. 
-// You must either update an existing threat model(s) to cover the changes made by this ADR, or add a new threat model.
+Admin API endpoints are protected by realm roles, hence only users that are assigned these roles can access the endpoints.
+
+No additional mitigation is required.
```
tombentley, line 62: I don't think the sentences added meet the requirements in the comments removed.

_adr/88/index.adoc
```diff
@@ -1,48 +1,71 @@
...
 
 ## Alternatives Considered / Rejected
 
 ## Challenges
-// What are the costs/drawbacks of the proposed decision?
```
tombentley, line 42: Surely there are some: Lots of Rover groups needed for large numbers of customers? IDK if this role membership information is imported into keycloak, or queried in-place in Rover, but if the latter then what are the consequences wrt availability?

tombentley (changes requested):

_adr/88/index.adoc
akoserwal: @tombentley What do you mean by "a large number of customers?". SRE group is the only consumer of these endpoints. Roles == rover groups, roles are created in keycloak. When a user request tokens, these roles are added as claims in tokens. Per service, we will have 6 roles (3 stage, 3 prod). If the number of services increases to 50, then roles will be 300; if the same set of users is added in all 300 groups, it increases the token size, impacting the admin endpoints workflow. Either request will get blocked at the gateway due to token size or time to parse/decode token will increase the response time.

tombentley: > SRE group is the only consumer of these endpoints.

That was the part I was missing, thanks.

But my point was really more general: Surely there are _some_ pros and cons to what's proposed?

_adr/88/index.adoc
pawelpaszki: right now it is 404 error (resource not found)

_adr/88/index.adoc
pawelpaszki: added a table. let me know if it makes more sense

_adr/88/index.adoc
pawelpaszki: done

_adr/88/index.adoc
pawelpaszki: added a note. does it make sense?

_adr/88/index.adoc
pawelpaszki: 100% agree that the concrete kas-fleet-manager threat model will need to be updated. I am reaching out to product security for more details on this process.

For this ADR, I understood the intent was not to link to Red Hat specific implementations since this ADR is generic, and there is no generic threat model that can be modified/attached, what is your opinion? @emmanuelbernard @tombentley

tombentley: @pawelpaszki that sounds reasonable. Can we add a sentence something like "This is a generic ADR so there is no threat model to update. The application of this ADR to specific services will require updating the threat model for those services"

_adr/88/index.adoc
pawelpaszki: @emmanuelbernard @tombentley - I have updated this section and created a document. please let me know if that's what you meant. I have also addressed all other comments, so I think this ADR is ready for a final review

tombentley (approved):

PR #82, opened 2022-10-17
feat: AP 17 Fleet shard status

shawkins:
This covers the pattern for fleet shard status that was discussed in an architecture call and several follow-up working group meetings.

The only major pending change is that @lburgazzoli would like to expand on the current vs desired state section to include a recommendation on using a generation field to know more precisely when the condition was observed.  That can be added after an initial merge if needed.

cc @tombentley

### DIFF
--- a/_ap/17/index.adoc
+++ b/_ap/17/index.adoc
@@ -1,0 +1,98 @@
+---
+num: 17
+category: "Fleet manager, fleet shard and instance operators"
+title: "Fleet shard status"
+status: "Draft"
+authors:
+  - "Steven Hawkins"
+tags:
+  - "kafka"
+---
+
+## Intent
+
+Define a pattern for fleet shard status representation and interpretation.
+
+## Motivation
+
+A fleet shard must convey the status of the data plane resources it is responsible for back to the control plane.  This is typically done using mechanisms built around the Kubernetes status sub-resource.  There should be consistency around the design and usage of that status information that conforms to Kuberentes guidelines and so that control planes, SREs, and developers can quickly and appropriately reason over the status. 
+
+## Status Guidelines
+
+### Conditions
+
+Kubernetes provides a significant amount of guidance for https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties[status conditions].
+
+Highlights:
+Condition types should be single word or short camel-cased names.  They should be unique - it’s more of a map of conditions by type, rather than just an array.
+The reason field should be a short word or short camel-cased phase.
+The message should be human readable.
+The status field should be a three-valued boolean.
+
+More specific to fleet shards:
+Since the fleet shard will typically send a full status object back to the control plane periodically or with every change, it is best to include only a minimally necessary set of conditions.  However that does not mean you can simply omit conditions.  If a condition is not present, it should be interpreted as having status Unknown.
+It is fine to have a control plane reason over status conditions - in particular the type, reason, and status fields.   However the parsing of a condition message should be avoided, and rather additional status fields should be used to provide specialized information.
+
+### Other Status Fields
+
+Beyond conditions, other status fields provide the ability to convey information via a custom schema.
+
+For example the ManagedKafka status once Ready will include route and other information, which are not part of a condition:
+
+[source,yaml]
+----
+status: 
+  conditions:
+    - lastTransitionTime: '2022-08-02T08:24:04.818994Z'
+      message: ''
+      status: 'True'
+      type: Ready
+  routes:
+    - name: admin-server
+      prefix: admin-server
+      router: ingresscontroller.kas.mk-0419-204008.t6kh.p1.openshiftapps.com
+    - name: bootstrap
+      prefix: ''
+      router: ingresscontroller.kas.mk-0419-204008.t6kh.p1.openshiftapps.com
+…  
+----
+
+### Interpretation
+
+The meaning of any condition or status field is up to the controller / resource and should be a documented part of the API contract.  
+
+For example the ManagedKafka status Ready condition refers to the operator’s view of the resource.  A ManagedKafka is only ready when all of the dependent resources are observed to be in their desired / ready state.  This will not always be the same notion of readiness from an end user’s observation of managed kafka service.  In particular, the restart of a broker pods, a temporary networking issue, etc. may not be reflected in ManagedKafka status.  It follows that a ManagedKafka status Ready condition alone is insufficient to show the user the state of their service.  For ManagedKafka canary and other kafka metrics provide a more exact representation of cluster functioning.
+Error States
+
+An operator in general may not consider any problem with a valid resource as terminal - one from which the resource can never recover.  The operator’s job is to implement the desired state whenever possible.  If at a given time additional resources are needed, other necessary system parts aren’t installed, etc.  - it doesn’t mean the conditions won’t be correct at a later time.  From the perspective of a managed service though we do have more rigid expectations.  For example a ManagedKafka should only be created with a Strimzi version that has an operator installed to manage.  For a fleet shard operator if such conditions are violated it is acceptable for the operator and the status handling to assume the resource is in a terminal state.
+
+When status is conveying an error that is not known to be terminal, it follows that the error may resolve on its own with additional time.  Control planes should assume such errors are ephemeral and not immediately react as if a hard failure has occurred.  If there are further actions that the control plane may take, the condition reason or other status fields should make that easy to determine.  This allows the data plane logic to remain straightforward and not have to fully understand every possible error condition.
+
+For example with managed kafka and auto-scaling enabled it may take longer for a node to become available than Strimzi’s timeout allows.  That error is propagated onto the ManagedKafka status.  However there is no further action expected by the control plane - and  with additional time the auto-scaled node will be available and the placement will succeed.  For example from https://issues.redhat.com/browse/MGDSTRM-9288 the control plane observed:
+
+[source,yaml]
+----
+status: 
+  conditions:
+    - lastTransitionTime: '2022-08-02T08:24:04.818994Z'
+      message: 'Exceeded timeout of 420000ms while waiting for Pods resource ninth-zookeeper-0 in namespace kafka-cbfv5rnfnecdu9rb4gc0 to be ready'
+      status: 'False'
+      reason: 'Error'
+      type: Ready
+----
+
+Rather than interpreting this as a hard failure, it should be assumed that with additional time and reconciliations that it will resolve on its own.
+
+How much time to give an error to resolve is up to the control plane and relevant SLOs.  During installation this may include waiting until the SLO has been exceeded.  For other operations it may mean waiting at least 1 additional resolution time window if the operator or dependent operators uses a time based mechanism.
+
+### Current vs. Desired State
+
+Especially in instances where spec changes are rolled into dependent resources the desired state will not be fully realized until some point in the future.  It is appropriate for the status to provide additional information about this transition.  As per the Kubernetes guidelines that could be represented via a condition, or similar to the standard Deployment resource additional status fields such as ready or available replicas, allow for the control plane or a user to infer the progress of the transition.
+Frequency Of Status Changes
+
+Rapid changes to a status should be avoided.  Each status change is an update that the local Kuberentes instance must process and is generally also relayed to the control plane, which can lead to a substantial amount of overhead for a large number of resources.  For controllers that implement fixed interval resolving this is generally not an issue.  Event driven controllers though should be designed to minimize unnecessary updates - in particular updates which do not imply a status change should leave the existing status unmodified.
+
+## Participants
+* Control Plane -- development team for the KAS Fleet Manager API.
+* Kafka Services -- team developing the kafka fleet shard operator.
+* MAS Connectors -- team developing the connector fleet shard operator.


lburgazzoli:
@shawkins will get back this week hopefully

tombentley:
@shawkins sorry to have dropped the ball so badly on this. Does it still need reviewing?

shawkins:
> @shawkins sorry to have dropped the ball so badly on this. Does it still need reviewing?

Yes it still needs to be reviewed.  There are some pending changes from @lburgazzoli that I'll make part of this pr as well.

shawkins:
@lburgazzoli incorporated the changes from the google doc, but omitted the part about polling based upon generation - that seems out of scope for this doc and would probably need to mention a watching paradigm in addition to polling.

@tombentley should be ready for review now.

_ap/17/index.adoc
```diff
@@ -0,0 +1,106 @@
...
+      router: ingresscontroller.kas.mk-0419-204008.t6kh.p1.openshiftapps.com
+…  
+----
+
+### Interpretation
+
+The meaning of any condition or status field is up to the controller / resource and should be a documented part of the API contract.  
+
+For example the ManagedKafka status Ready condition refers to the operator’s view of the resource.  A ManagedKafka is only ready when all of the dependent resources are observed to be in their desired / ready state.  This will not always be the same notion of readiness from an end user’s observation of managed kafka service.  In particular, the restart of a broker pods, a temporary networking issue, etc. may not be reflected in ManagedKafka status.  It follows that a ManagedKafka status Ready condition alone is insufficient to show the user the state of their service.  For ManagedKafka canary and other kafka metrics provide a more exact representation of cluster functioning.
+Error States
```
tombentley, line 66: Should this be a heading?

_ap/17/index.adoc
```diff
@@ -0,0 +1,106 @@
...
+
+An operator in general may not consider any problem with a valid resource as terminal - one from which the resource can never recover.  The operator’s job is to implement the desired state whenever possible.  If at a given time additional resources are needed, other necessary system parts aren’t installed, etc.  - it doesn’t mean the conditions won’t be correct at a later time.  From the perspective of a managed service though we do have more rigid expectations.  For example a ManagedKafka should only be created with a Strimzi version that has an operator installed to manage.  For a fleet shard operator if such conditions are violated it is acceptable for the operator and the status handling to assume the resource is in a terminal state.
```
tombentley, line 68: I understand what this is trying to say, but I think that way this is worded is a little vague. Do you think the following is better?

> An operator in a fleetshard _usually_ lacks sufficient context to determine whether a valid resource is in a terminal state. The operator’s job is simply to implement the desired state whenever possible.  For example, if at a given time additional resources are needed, other necessary system parts aren’t installed, etc.  - it doesn’t mean the conditions won’t be correct at a later time as a result of some action outside the context of the operator. 
> 
> In cases where the operator does have sufficient context to determine terminality, for example violation of preconditions in installed versions of APIs, then it is acceptable for the operator and the status handling to assume the resource is in a terminal state.

_ap/17/index.adoc
```diff
@@ -0,0 +1,106 @@
...
+
+Especially in instances where spec changes are rolled into dependent resources the desired state will not be fully realized until some point in the future.  It is appropriate for the status to provide additional information about this transition.  As per the Kubernetes guidelines that could be represented via a condition, or similar to the standard Deployment resource additional status fields such as ready or available replicas, allow for the control plane or a user to infer the progress of the transition.
+
+### generation and observedGeneration
+
+In kubernetes, each object should have a generation field which is defined as a sequence number - set by the system and monotonically increased per resource on a spec change. Some resources include a field named observedGeneration as part of the status, which is the generation most recently observed by the component responsible for acting upon changes to the desired state of the resource. This can be used, for instance, to ensure that the reported status reflects the most recent desired state. The observedGeneration field is also part of metav1.Conditions and represents the spec generation that the condition was set based upon.
```
tombentley, line 95:
```suggestion
In kubernetes, each object should have a `metadata.generation` field which is defined as a sequence number - set by the system and monotonically increased per resource on a spec change. Some resources include a field named `status.observedGeneration`, which is the generation most recently observed by the component responsible for acting upon changes to the desired state of the resource. This can be used, for instance, to ensure that the reported status reflects the most recent desired state. The observedGeneration field is also part of metav1.Conditions and represents the spec generation that the condition was set based upon.
```

tombentley (approved):
I left a few suggestions, but overall this LGTM. Thanks!

_ap/17/index.adoc
```diff
@@ -0,0 +1,106 @@
+---
+num: 17
+category: "Fleet manager, fleet shard and instance operators"
+title: "Fleet shard status"
+status: "Draft"
```
tombentley, line 5: Let's assume this will be accepted.

```suggestion
status: "Accepted"
```

_ap/17/index.adoc
```diff
@@ -0,0 +1,106 @@
...
+    - lastTransitionTime: '2022-08-02T08:24:04.818994Z'
+      message: 'Exceeded timeout of 420000ms while waiting for Pods resource ninth-zookeeper-0 in namespace kafka-cbfv5rnfnecdu9rb4gc0 to be ready'
+      status: 'False'
+      reason: 'Error'
+      type: Ready
+----
+
+Rather than interpreting this as a hard failure, it should be assumed that with additional time and reconciliations that it will resolve on its own.
+
+How much time to give an error to resolve is up to the control plane and relevant SLOs.  During installation this may include waiting until the SLO has been exceeded.  For other operations it may mean waiting at least 1 additional resolution time window if the operator or dependent operators uses a time based mechanism.
```
maleck13, line 87: how much time before the control plane reports the error or considers it a terminal error? Reading this, I would assume the control plane is aware of an error state having occurred and is using other metrics (time) to decide when it feels the error state has become exceptional and something  (beyond waiting) may need to be done?

_ap/17/index.adoc
```diff
@@ -0,0 +1,106 @@
...
+status: 
+  conditions:
+    - lastTransitionTime: '2022-08-02T08:24:04.818994Z'
+      message: 'Exceeded timeout of 420000ms while waiting for Pods resource ninth-zookeeper-0 in namespace kafka-cbfv5rnfnecdu9rb4gc0 to be ready'
+      status: 'False'
+      reason: 'Error'
+      type: Ready
+----
+
+Rather than interpreting this as a hard failure, it should be assumed that with additional time and reconciliations that it will resolve on its own.
```
maleck13, line 85: Perhaps this should be `may` resolve on its own?

maleck13 (approved):

_ap/17/index.adoc
shawkins: Incorporated the above, please see if that matches what you are thinking.

_ap/17/index.adoc
```diff
@@ -0,0 +1,106 @@
...
+
+How much time to give an error to resolve is up to the control plane and relevant SLOs.  During installation this may include waiting until the SLO has been exceeded.  For other operations it may mean waiting at least 1 additional resolution time window if the operator or dependent operators uses a time based mechanism.
+
+### Current vs. Desired State
+
+Especially in instances where spec changes are rolled into dependent resources the desired state will not be fully realized until some point in the future.  It is appropriate for the status to provide additional information about this transition.  As per the Kubernetes guidelines that could be represented via a condition, or similar to the standard Deployment resource additional status fields such as ready or available replicas, allow for the control plane or a user to infer the progress of the transition.
```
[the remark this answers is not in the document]
shawkins: Elaborated more based upon a kafka version upgrade.

EricWittmann (approved):

lburgazzoli (commented):
LGTM (I'm among the authors so I should not be among the reviewer)

shawkins:
@emmanuelbernard @danielezonca please review or delegate so that this can move forward

_ap/17/index.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+Similar concepts with fields that mirror their kubernetes couterparts can be introduced to MAS resources to offer a standard way for clients and control planes to figure out when a specific desired state has been taken into account. 
+
+### Frequency Of Status Changes
+
+Rapid changes to a status should be avoided.  Each status change is an update that the local Kuberentes instance must process and is generally also relayed to the control plane, which can lead to a substantial amount of overhead for a large number of resources.  For controllers that implement fixed interval resolving this is generally not an issue.  Event driven controllers though should be designed to minimize unnecessary updates - in particular updates which do not imply a status change should leave the existing status unmodified.
```
danielezonca, line 106: Can you please clarify this scenario? 
If I have understood correctly, this refers to scenario like there are multiple iteration that change "something" in the CR but the overall status is going to remain "creating" or similar.
Is it correct?

In general I understand the principle and I agree but in some scenarios this might be necessary to avoid "implicit contracts" based on other fields

danielezonca (approved):
LGTM, just a minor comment

_ap/17/index.adoc
shawkins: No this is not based upon CR changes. This is warning against trying to make the status too fine grained / immediate - suppose someone proposed reporting a health probe of every pod that is transitively under a ManagedKafka on the ManagedKafka status.  Presuming that include the timestamp of the latest probe you could be looking at a modification to the status every couple of seconds, which would generate an event locally that needs to be processed as well as need to be relayed to the control plane.  That would not be advisable.

tombentley:
@emmanuelbernard still waiting for your review on this one.

emmanuelbernard (changes requested):
This is generally useful thank you for the write up.
I felt the description was assuming a good knowledge of fleet shard and connecting the dot required to me some leap of my imagination.
That's why I've been heavily proposing to add lots of example.

Generally I think all recommendations should try and succintly explan
- why it's useful (value)
- what for (context)
- how it is done (with examples)

The current structure of the document reflects the CR areas and I wonder whether a reorganisation around concrete problems could highlight its usefulness.

_ap/17/index.adoc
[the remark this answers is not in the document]
shawkins: Added the yaml that matches the description.

_ap/17/index.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+For example the upgrade of ManagedKafka kafka version requires a rolling update of brokers where each broker is taken off-line and replace with one that has an updated version.  This is a relatively long-running process.  The status ManagedKafka Ready condition during this time will have a reason of KafkaUpdating, rather than simply indicating Ready based upon the Strimzi Kafka resource being updated to the desired state.
+
+### generation and observedGeneration
+
+In kubernetes, each object should have a `metadata.generation` field which is defined as a sequence number - set by the system and monotonically increased per resource on a spec change. Some resources include a field named `status.observedGeneration`, which is the generation most recently observed by the component responsible for acting upon changes to the desired state of the resource. This can be used, for instance, to ensure that the reported status reflects the most recent desired state. The observedGeneration field is also part of metav1.Conditions and represents the spec generation that the condition was set based upon.
```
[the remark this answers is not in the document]
shawkins: Added an example.

_ap/17/index.adoc
```diff
@@ -0,0 +1,111 @@
...
+  - "Luca Burgazzoli"
+tags:
+  - "kafka"
+---
+
+## Intent
+
+Define a pattern for fleet shard status representation and interpretation.
+
+## Motivation
```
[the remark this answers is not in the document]
shawkins: After starting on a drawing it made me wonder if that was making this seem too complicated.  So I just updated with some language about custom resources in general - I really am trying to convey this is the kubernetes custom resource paradigm with an intermediary, so it has some additional considerations on top of the base recommendations.

_ap/17/index.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+Highlights:
+Condition types should be single word or short camel-cased names.  They should be unique - it’s more of a map of conditions by type, rather than just an array.
+The reason field should be a short word or short camel-cased phase.
+The message should be human readable.
+The status field should be a three-valued boolean.
+
+More specific to fleet shards:
+Since the fleet shard will typically send a full status object back to the control plane periodically or with every change, it is best to include only a minimally necessary set of conditions.  However that does not mean you can simply omit conditions.  If a condition is not present, it should be interpreted as having status Unknown.
+It is fine to have a control plane reason over status conditions - in particular the type, reason, and status fields.   However the parsing of a condition message should be avoided, and rather additional status fields should be used to provide specialized information.
```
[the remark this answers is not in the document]
shawkins: Are there specific examples you had in mind?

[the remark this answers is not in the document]
emmanuelbernard: Things like showing non omited considtions even though not use for that specific effective status report.
And that might not be here but you say that the shard in most case cannot know whether a case is transient or terminal. which makes me wonder how the control plan makes that decision from thge status it receives fromt he shard. HEre some examples of how you achieved it , or how a status is interpreted for feedback tot he end user would be interesting for context on how the global machinery is orchestrated

_ap/17/index.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+### Error States
+
+An operator in a fleetshard _usually_ lacks sufficient context to determine whether a valid resource is in a terminal state. The operator’s job is simply to implement the desired state whenever possible. For example, if at a given time additional resources are needed, other necessary system parts aren’t installed, etc. - it doesn’t mean the conditions won’t be correct at a later time as a result of some action outside the context of the operator.
+
+In cases where the operator does have sufficient context to determine terminality, for example violation of preconditions in installed versions of APIs, then it is acceptable for the operator and the status handling to assume the resource is in a terminal state.
```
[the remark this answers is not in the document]
shawkins: Made some further refinements to the text and added an example.

_ap/17/index.adoc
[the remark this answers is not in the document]
shawkins: > Things like showing non omited considtions even though not use for that specific effective status report.

The guidance about minimizing is fleetshard specific.  The interpretation of missing as unknown comes straight from the kubernetes recommendations.  From existing non-fleetshard operators, we have counter examples like:

strimzi uses both a Ready and NotReady conditions.  The NotReady condition is only populated when there's a problem.  By their contract you have to know that a missing NotReady means to look instead for the Ready condition, not that NotReady is currently unknown.

But I'm honestly not sure how instructive that is.

> which makes me wonder how the control plan makes that decision from thge status it receives fromt he shard

That's the gist of the recommendations around the error handling - the control plane nor a user can assume an immediate action is needed given seeing something is currently in error - regardless of whether we're installing.

The fleetshard operator interprets the strimzi status as follows: 
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/blob/main/operator/src/main/java/org/bf2/operator/operands/AbstractKafkaCluster.java#L169  That is then aggregated with the other operands.  The fleet manager is further able to aggregate reasoning over the cluster itself and from things like route53.  

We don't expect the fleet manager to make specific decisions based upon errors that appear in our status - we are instead calling out very specific service level cases - the wrong profile or version of something is in the cr, the data plane lacks capacity for the given instance - that the control plane may use to do things like select a different strimzi version or use a different cluster.  Otherwise about the best that they can infer is whether we think we're still Installing.

PR #84, opened 2022-12-08
feat: add an artchitecture pattern for certificate management

machi1990:
This depends on https://github.com/bf2fc6cc711aee1a0c2a/architecture/pull/83


emmanuelbernard:
hey @machi1990, happy new year :)
I wanted to highlight to you that we are using a template for APs
https://architecture.appservices.tech/ap/0/
Might be good to use it early on.

machi1990:
Hey @emmanuelbernard bonne année :-)

Thanks for the pointers. I've it on my TODO list (after long PTO) and will get to it eventually.

machi1990:
> Hey @emmanuelbernard bonne année :-)
> 
> Thanks for the pointers. I've it on my TODO list (after long PTO) and will get to it eventually.

Done. This is ready for another round of review @emmanuelbernard

_ap/17/index.adoc
```diff
@@ -0,0 +1,62 @@
...
+  - "Manyanda Chitimbo"
+---
+
+## Intent
+
+Define a pattern for the automatic management of certificates per data plane cluster using Letsencrypt
+
+## Motivation
+
+When a managed service is deployed on the data plane cluster its routes have be reachable via TLS.
```
emmanuelbernard, line 20:
```suggestion
When a managed service is deployed on the data plane cluster its routes have to be reachable via TLS.
```

_ap/17/index.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+It means that a managed service a way to the clients to verify and validate the TLS connection with the related handshake with the service exposed outside of the data plane cluster.
```
emmanuelbernard, line 22: I did not understand this sentence, the beginning.

_ap/17/index.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+## Applicability
+
+This pattern applies to any service that wants a managed service routes to be reachable via TLS
```
emmanuelbernard, line 26:
```suggestion
This pattern applies to any service deployed on a dataplane that wants a managed service routes to be reachable via TLS
```

_ap/17/index.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+## Description
+
+A managed service will use https://Letsencrypt.org[Letsencrypt] to managed certificates on the fly i.e automatic and this removing the need for manual intervention and the toil related to this.
```
emmanuelbernard, line 30:
```suggestion
A managed service will use https://Letsencrypt.org[Letsencrypt] to managed certificates on the fly i.e automatic and thus removing the need for manual intervention and the toil related to this.
```

_ap/17/index.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+To avoid sharing of certificates between data planes, managed service implementors are recommended to generate a wildcard certificate per data plane.
+
+As an example, if the root domain is `myservice.bf2.dev`, then a wildcard certificate will be generated for each data plane cluster under a subdomain `<cluster-id>.myservice.bf2.dev`.
+
+Furthermore, all routes for the managed services deployed on the the data plane will be of the form `foobar.<cluster-id>.myservice.bf2.dev`.
+
+NOTE: There is a more granular approach to avoid certificate sharing by generating a widlcard certificate per deployed service, however this approach is likely to encounter and hits https://Letsencrypt.org/docs/rate-limits/[Letsencrypt limits] more frequently.
```
emmanuelbernard, line 38:
```suggestion
NOTE: There is a more granular approach to avoid certificate sharing by generating a widlcard certificate per deployed service instance, however this approach is likely to encounter and hits https://Letsencrypt.org/docs/rate-limits/[Letsencrypt limits] more frequently. 
```

_ap/17/index.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+This pattern applies to any service that wants a managed service routes to be reachable via TLS
+
+## Description
+
+A managed service will use https://Letsencrypt.org[Letsencrypt] to managed certificates on the fly i.e automatic and this removing the need for manual intervention and the toil related to this.
+
+To avoid sharing of certificates between data planes, managed service implementors are recommended to generate a wildcard certificate per data plane.
+
+As an example, if the root domain is `myservice.bf2.dev`, then a wildcard certificate will be generated for each data plane cluster under a subdomain `<cluster-id>.myservice.bf2.dev`.
```
emmanuelbernard, line 34: I forgot whether and what we said about exposing the cluster id in the URL. It does contradict SDR 1. I'll add it tot he conversastion in the arch call.

emmanuelbernard: BTW it is unclear what cluster id refers to. Make it more explicit (e.g. Kubernetes cluster id). I thought it was the kafka cluster id for a few minutes until I chcked ADR-89

emmanuelbernard (changes requested):

_ap/17/index.adoc
machi1990: > I forgot whether and what we said about exposing the cluster id in the URL. It does contradict SDR 1. I'll add it tot he conversastion in the arch call.

> BTW it is unclear what cluster id refers to. Make it more explicit (e.g. Kubernetes cluster id). I thought it was the kafka cluster id for a few minutes until I chcked ADR-89

Thanks for raising it. Non, it is not the Kafka id but the data plane cluster id as we need a wildcard cert per data plane cluster.

I'll make it clear.

_ap/17/index.adoc
machi1990: Ah reading this again and I feel like something might have slipped out before I pushed the changes. I'll revisit this.
Thanks for the review.

_ap/17/index.adoc
machi1990: > I forgot whether and what we said about exposing the cluster id in the URL. It does contradict SDR 1. I'll add it tot he conversastion in the arch call.

Thanks, we are using the cluster id for now in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1504 but we could as well envision to generate a unique slug per cluster and use it in the URL. 
@emmanuelbernard do you think we'll get a reply from the call soonish enough for us to go ahead with the change in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1504? Thanks

/cc @pb82

machi1990: @emmanuelbernard permit me to bring a bit of precision on 

> Thanks, we are using the cluster id for now in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1504 but we could as well envision to generate a unique slug per cluster and use it in the URL.

We need a unique identifier for the shard and the `OCM` cluster id was choosen since its unique. 
For example for the `kafka.bf2.dev` domain, and the cluster id `123`, then the base domain for this cluster will be `123.kafka.bf2.dev` and we'll generate a wildcard certificate for it `*.123.kafka.bf2.dev`

When kafkas are deployed on this data plane cluster the kafka urls will be of the form `xxxxx.123.kafka.bf2.dev` e.g the admin server url can be `admin-server-kafka-cf-ho-a---r-kce--otg` 

Going back to what I said above

> We need a unique identifier for the shard and the `OCM` cluster id was choosen since its unique. 

If we deem that the cluster-id shouldn't be used as the shard1, we could envision to generate a unique string that can be used as a shard id. We could use the logical cluster id (the PK in KFM database for the given cluster) or we could generate a totally different shard id that's unique and totally opaque without any meaning. 

Let me know what you think and I'll update the AP and reflects the changes in the ADR-89.

emmanuelbernard: The opaque proposal does not solve the core reason that the URL is no longer location agnostic at least if we make the assumption that the certificate remains only on a given data plane cluster and does not travel.

machi1990: Thanks. I think I see the issue now e.g when a kafka is migrated from one OSD to another...

machi1990:
Hi @pb82 @emmanuelbernard I've revisited the architecture pattern following recent Architecture discussions.
I've created a new ADR-90 to supersede [ADR-89 - certificate per data plane](https://github.com/bf2fc6cc711aee1a0c2a/architecture/blob/main/_adr/89/index.adoc)
Please have a look.

_adr/90/index.adoc
```diff
@@ -0,0 +1,135 @@
...
+The link:../89/index.adoc[ADR-89: A bf2 subdomain & wildcard certificate per data plane cluster] proposed a solution of automatic certificate management per data plane cluster, the solution however made the Kafka URLs non portable i.e the URLs are non deployment location agnostic.
+URLs portability is the design intention we do not want to loose - as it allows migrating a situation where the need to migrate a Kafka from one data plane to another becoming only an infrastructure change that the consumer of the Kafka won't need to be soo much concerned about by updating the various kafka URLs (admin server API & bootstrap server host)
+
+## Proposal
+// What is the decision being proposed
+The proposed solution is going to use https://Letsencrypt.org[Letsencrypt] to automatically manage certificates.
+The solution consists of: 
+
+1. Generating one wildcard certificate per Kafka instance .
+For each new `standard` Kafka instance, there will be a wildcard certificate for the `*.<kafka-id>.kafka.bf2.dev` subdomain where `<cluster-id>` is unique for each cluster.
```
pb82, line 52: cluster-id -> kafka-id

pb82 (approved):
@emmanuelbernard are you ok with accepting the updated proposal?

_adr/90/index.adoc
machi1990: good catch!

pb82 (approved):

tombentley:
@emmanuelbernard could you do another pass?

_ap/17/index.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+## Applicability
+
+This pattern applies to any service deployed on a dataplane that wants a managed service routes to be reachable via TLS
+
+## Description
+
+A managed service will use https://Letsencrypt.org[Letsencrypt] to managed certificates on the fly i.e automatic and thus removing the need for manual intervention and the toil related to this.
+
+To avoid sharing of certificates between deployed instances, managed service implementors are recommended to generate a wildcard certificate per managed instance.
```
maleck13, line 31: is it not wildcard or dedicated certificate?

_ap/17/index.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+As an example, if the root domain is `myservice.bf2.dev`, then a wildcard certificate will be generated for each new managed instance under a subdomain `<managed-instance-id>.myservice.bf2.dev`. The `<managed-instance-id>` is the unique id of the managed instance.
+
+Furthermore, all routes for the managed services deployed on the the data plane will be of the form `<unique-route-slug>.<managed-instance-id>.myservice.bf2.dev`.
+If we the Managed Kafka Service as an example, it will expose Kafka routes with the form `<unique-route-slug>.<kafka-instance-id>.kafka.bf2.dev` e.g `boostrapserverhost.123456.kafka.bf2.dev`.
+
```
maleck13, line 37: should we qualify as in if there is a possibility of using multiple subdomains as hosts under a given instance id then a wildcard per instance id is recommended. If there is only ever a single subdomain then there is no need for a wildcard. I guess what I am saying is I am not sure we should force the use of a wildcard if a single domain cert would suffice

maleck13 (commented):
General LGTM, I have a question on the AP. Not sure we should always have to use a wildcard

_ap/17/index.adoc
machi1990: The suggestion make sense. I can reword it in a way you are suggesting.

machi1990:
> General LGTM, I have a question on the AP. Not sure we should always have to use a wildcard

Thank you. I'll update the AP to mention other possibilities e.g a certificate per route of a managed instance.

_ap/17/index.adoc
machi1990: The AP was written in a way that there will be a dedicated wildcard certificate for all routes of a managed instance.
For example,  if a managed instance has 5 routes, all those routes will be using one dedicated wildcard certificate specific to that instance. This is the case of Kafkas for example.

Another possibility and even more fine grained approach is to have a dedicated certificate per route. Which is what I think you are suggesting to include in https://github.com/bf2fc6cc711aee1a0c2a/architecture/pull/84#discussion_r1109437233 - I'll update the AP to mention this as well.

_ap/17/index.adoc
```diff
@@ -0,0 +1,59 @@
...
+A managed service will use https://Letsencrypt.org[Letsencrypt] to managed certificates on the fly i.e automatic and thus removing the need for manual intervention and the toil related to this.
+
+To avoid sharing of certificates between deployed instances, managed service implementors are recommended to generate a dedicated certificate per managed instance.
+
+As an example, if the root domain is `myservice.bf2.dev`, then a dedicated certificate will be generated for each new managed instance under a subdomain `<managed-instance-id>.myservice.bf2.dev`. The `<managed-instance-id>` is the unique id of the managed instance.
+
+Furthermore, all routes for the managed services deployed on the the data plane will be of the form `<managed-instance-id>.myservice.bf2.dev`.
+If we take the Managed Kafka Service as an example, it generates a dedicated wildcard certificate per managed Kafka instance - see link:../../adr/90[ADR-90].
+All the Kafka routes are of the form `<unique-route-slug>.<kafka-instance-id>.kafka.bf2.dev` e.g `boostrapserverhost.123456.kafka.bf2.dev`.
```
tombentley, line 37: It feels like this section should explicitly call out that the subdomain used for an instance must be independent of the dataplane in which it (currently) resides.

_ap/17/index.adoc
```diff
@@ -0,0 +1,59 @@
...
+What this pattern brings:
+
+* Automated management of certificates per managed instance.
+* Avoid sharing of certificate, if one certificate is compromised, the rest of the certiicates are not affected
+* Consistency with other services which also apply this AP
+
+However there are challenges that service implementors have to be aware of:
+
+* Once the certificates are generated, the managed service have to secure storing of the auto generated certificate in Vault and ensuring that only priveledged users e.g only SRE, have access to Vault.
+* Once the certificates in the Data Plane, they should be in a Kubernetes secret that only priveldged users can see.
```
tombentley, line 57:
```suggestion
* Once the certificates in the Data Plane, they should be in a Kubernetes secret that only privelidged users can see. 
```

_adr/90/index.adoc
```diff
@@ -0,0 +1,135 @@
...
+The intention is to reduce the risk of compromising all our certificates in case there is an outage or an issue with the renewal system in Letsencrypt. 
+It also removes the chances of bombarding Letsencrypt with bulk requests or rollout of all the Kafkas in one big batch after renewals. 
+
+5. Revocation will be handled per Kafka instance.
+It will follow a GitOps approach where SRE will request to revoke a certificate for a Kafka.
+Once the compromised certificate has been revoked, a new certificate for that Kafka will be generated.
+For already existing Kafkas, they'll continue to share the newly generated certificate: see proposal number (3). 
+
+6. When a standard Kafka is removed from the fleet manager, its corresponding certificate will also be revoked. 
+This is with an excpetion for all Kafkas that uses a shared certificate i.e Kafkas that existed before the solution is in place or all new developer Kafkas that uses a shared wildcard `*.trial.kafka.dev` certificate.
```
tombentley, line 80:
```suggestion
This is with an exception for all Kafkas that uses a shared certificate i.e Kafkas that existed before the solution is in place or all new developer Kafkas that uses a shared wildcard `*.trial.kafka.dev` certificate.
```

_adr/90/index.adoc
```diff
@@ -0,0 +1,135 @@
...
+
+7. There will be separate Letsencrypt's account per kas-fleet-manager environment. 
+For each of these acccounts, a request to overrides the various https://Letsencrypt.org/docs/rate-limits[Letsencrypt limits] will be done by following https://Letsencrypt.org/docs/rate-limits/#a-id-overrides-a-overrides[Letsencrypt limits overrides].
+There are two main limits that we should be aware of and are of interests: (1) Certificates per Registered Domain which has a default of 50 certificates per week (2) You can create a maximum of 300 New Orders per account per 3 hours.
+Each of these two limits needs to be bumped to a bigger value after capacity planning for certificate generation has been done.
+
+NOTE: There is an existing logic in the data plane that ensures that all sensitive parts in the ManagedKafka CR are not written in plain text. 
+Instead a secret is created and only references to the secrets will be added before the ManagedKafka is created in the data plane.
+
```
tombentley, line 89: Does the dataplane logic already handle renewal and revocation? If so, let's say so explicitly.

In particular, what has the logic for knowing about the impending passing of a certificate's `notAfter` (bearing in mind that the certificate in question might be a root or intermediate CA cert, though I'd expect Let's Encrypt would be avoiding issuing end entity certificates with `notAfter` after any CA certificate in the chain).

_adr/90/index.adoc
```diff
@@ -0,0 +1,135 @@
...
+### Threat model
+// Provide a link to the relevant threat model. 
+// You must either update an existing threat model(s) to cover the changes made by this ADR, or add a new threat model.
+
+Once the certificate are in the data plane in form of a Kubernetes secret, a priviledged users can still see them.
```
tombentley, line 94: Is the access to these `Secret`s already controlled using RBAC? If so, let's say so explicitly.

tombentley (approved):
Thanks. I leve a few nits, but overall LGTM.

_ap/17/index.adoc
machi1990: Thanks. I'll add that in.

_adr/90/index.adoc
machi1990: For Managed Kafka, the data plane components are only responsible for stripping out the certificate from the MK CR and creating a secret out of it.
Certificate management is handled entirely in the fleet manager and it is responsible for automatic generation, renewal (within 1 month before the certificate expires and revocation - when a managed instance is removed or via an admin endpoint, once we detect that there are certificates that have been compromised)

_adr/90/index.adoc
machi1990: Yes. That's already the case for MK. 
It is called out in the line below 
```
For non enterprise data plane clusters, this is mitigated by only allowing SREs to have the priviledge / permissions to see these secrets.
```

machi1990:
Hey folks, the review comments have been addressed. Bumping this for further review if needed or merging it. Thanks

EricWittmann (approved):
lgtm

_adr/90/index.adoc
```diff
@@ -0,0 +1,136 @@
...
+- Avoid exposing domain registrar & certificate issuers credentials onto enterprise data plane clusters 
+
+## Non-goals
+// Bulleted list of outcomes that this ADR is not trying to achieve.
+- Creating a pattern for and/or developing a central certificate management service
+
+## Current situation
+// Where are we now?
+
+The https://architecture.appservices.tech/adr/5/[ADR-5: Kafka Cluster Certificate Generation] describes the current situation where one certificate is shared across data planes. The certificate is renewed manually using the https://github.com/bf2fc6cc711aee1a0c2a/kas-sre-sops/blob/main/sops/key_management/control_plane/dataplane_certificate.asciidoc[data plane certificate SOP]
```
danielezonca, line 41: SOP link is to a private repo

_adr/90/index.adoc
```diff
@@ -0,0 +1,136 @@
...
+For enterprise data plane clusters, customers will be advised to do so as well.
+Morever, since there will be a different certificate per kafka instance, if one certificate is compromised, it won't affect others and it can be safely revoked. 
+The proposed solution also limits the possibility of an owner of a enterprise data plane cluster, eavesdropping on another data plane cluster that they do not own as the certificate keys will be different. 
+
+NOTE: If the shared certificate for an existing Kafka is to be compromised, then it opens up a securiry risk on all existing Kafkas. However, the chances of this security breach occurring is mitigated by first and foremost ensuring only SREs have privileged access to data plane clusters. And second, if a the security issue is detected then certificates can be easily revoked and new ones autocreated 
+
+## Alternatives Considered / Rejected
+
+1. Deploying https://www.redhat.com/sysadmin/cert-manager-operator-openshift[cert-manager operator] to the data plane. Rejected to avoid exposing domain registrar credentials and to avoid adding another component that consumes data plane resources and that needs to be monitored.
```
danielezonca, line 106: I'm not an expert of this operator but does it rely on some external system or is it selfcontained/standalone?

I'm asking because one drawback of the current proposal is that Letsencrypt is going to be a SPOF for the service (at least during the creation of the instance) and we will probably need to consider in the future some sort of fallback to mitigate this risk

_adr/90/index.adoc
```diff
@@ -0,0 +1,136 @@
...
+All standard Kafkas will have a wildcard certificate on the `*.<kafka-id>.kafka.bf2.dev` subdomain where the `kafka-id` is unique for each Kafka.
+All new developer Kafkas (short living instances of 48 hours) will have a wildcard certificate on the `*.trial.kafka.bf2.dev` subdomain, and they will share the same wildcard certificate. This is done so as to keep our usage of the https://Letsencrypt.org/[Letsencrypt] sane and avoid frequently reaching their https://Letsencrypt.org/docs/rate-limits/[limits] due to spikes in creation of developer kafkas.
+2. Reaching the various advertised https://Letsencrypt.org/docs/rate-limits/[Letsencrypt limits] is also a challenge.
+However, the various limits once reached can be overridden via a request as described in https://Letsencrypt.org/docs/rate-limits/#a-id-overrides-a-overrides[Letsencrypt limits overrides]. The usage of the https://Letsencrypt.org/[Letsencrypt] is kept sane by sharing an auto managed wildcard certificate for all new developer Kafkas.
+
+## Dependencies
+// What are the knock-on effects if this decision is accepted?
+
+Depedency on Letsencyrpt and its limits
```
danielezonca, line 130: I would expect part of the limits of Letsencyrpt can be solved with some enterprise/paid plan.
Am I correct? If so I think we can mention that limits are tied to the current tied that we use

_ap/17/index.adoc
```diff
@@ -0,0 +1,58 @@
+---
+num: 17
+tags:
+  - "tls"
+  - "certificates"
+  - "dns"
+  - "Letsencrypt"
+title: "Automatic management of certificates per managed instance using Letsencrypt"
```
danielezonca, line 8: How much of this can/is/make sense to include in factorised fleet manager? 
Is it possible to make it "almost" for free for new services?

danielezonca (approved):
Few comments but overall LGTM 👍

_ap/17/index.adoc
machi1990: There are ACME libraries / operators that can be used. That's what I'll advise new services using depending on their deployment topology. 
Also the way this is implemented in kafka services, can be extracted onto its own SDK as part of the factorized fleet management effort. The SDK part is the second phase of that effort and involves multiple part of the the KFM codebase.

_adr/90/index.adoc
machi1990: No, Letsencrypt doesn't have the notion of enterprise plan as far as I am aware. However, they've the option to request limits override which has been mentioned in this ADR. The process needs to be followed and they (letsencrypt) will get intouch if further clarification is needed.

Other services like zerossl have the notion of plan - and they charge a fee for this.

_adr/90/index.adoc
machi1990: Yes, we do not have a public / open source friendly SOP - so I guess we'll keep the link as it is for now until we've an equivalent.

_adr/90/index.adoc
machi1990: I feel like the comment goes beyond the operator which I do not have much experience on but on how reliable Letsencrypt can be on how usage around it. So +1 to mitigate the risk of it becoming a SPOF - e.g by building retry mechanisms (some acme libraries have this) so that our SLOs are not impacted much.

_adr/90/index.adoc
danielezonca: Ok thanks for the clarification, maybe we can add a note/comment to say that "private link"

emmanuelbernard (approved):

tombentley:
@lburgazzoli please can you review?

tombentley:
Also @maleck13 are you happy with this now?

machi1990:
@maleck13 @lburgazzoli can you've a look? Thanks

machi1990:
@maleck13 @lburgazzoli friendly ping to have a look at this PR. I'd like to merge this soon if possible, thanks.
/cc @emmanuelbernard @tombentley

_adr/90/index.adoc
```diff
@@ -0,0 +1,136 @@
...
+status: "Accepted" # One of Draft, Accepted, Rejected
+authors:
+  - "Manyanda Chitimbo"
+tags:
+  - "kafka" # e.g. kafka, connectors, registry
+applies_padrs: # What PADRs does this ADR apply?
+applies_patterns: # What APs does this ADR apply?
+---
+
+// Top style tips:
```
maleck13, line 13: should all these commented out bits be here?

_adr/90/index.adoc
```diff
@@ -0,0 +1,136 @@
...
+
+## Current situation
+// Where are we now?
+
+The https://architecture.appservices.tech/adr/5/[ADR-5: Kafka Cluster Certificate Generation] describes the current situation where one certificate is shared across data planes. The certificate is renewed manually using the https://github.com/bf2fc6cc711aee1a0c2a/kas-sre-sops/blob/main/sops/key_management/control_plane/dataplane_certificate.asciidoc[data plane certificate standard operationg procedure (SOP)]: the SOP link points to a private repository and requires membership permission to see it..
+
+The link:../89/index.adoc[ADR-89: A bf2 subdomain & wildcard certificate per data plane cluster] proposed a solution of automatic certificate management per data plane cluster, the solution however made the Kafka URLs non portable i.e the URLs are non deployment location agnostic.
+URLs portability is the design intention we do not want to loose - as it allows migrating a Kafka from one data plane to another becoming only an infrastructure change that the consumer of the Kafka won't need to be soo much concerned about and be required to update the various kafka URLs (admin server API & bootstrap server host).
```
maleck13, line 44: soo -> so

machi1990: good catch. Updated

_adr/90/index.adoc
```diff
@@ -0,0 +1,124 @@
...
+Existing Kafkas will continue to share the wildcard certificate on the `*.kafka.bf2.dev` domain.  
+The shared certificate for the already existing Kafkas will now be automatically renewed by the fleet manager.
+
+4. Certificate renewals will also be handled automatically by the fleet manager using the https://github.com/caddyserver/certmagic[certmagic library]. 
+Renewals will be blast radius aware to renew certificates in small runs on per Kafka basis.
+Renewals of each certificate will happen 30 days before the certificate expires. 
+The intention is to reduce the risk of compromising all our certificates in case there is an outage or an issue with the renewal system in Letsencrypt. 
+It also removes the chances of bombarding Letsencrypt with bulk requests or rollout of all the Kafkas in one big batch after renewals. 
+
+5. Revocation will be handled per Kafka instance.
```
maleck13, line 67: revocation is an interesting area. As I understand it, it relies on the client actually checking if it has been revoked. From some brief reading, this seems to often not be the case. https://github.com/cert-manager/cert-manager/issues/392 . Below seems to describe replacing a compromised cert rather than actively revoking it and adding it to the LE revoked cert list? https://letsencrypt.org/docs/revoking/

_adr/90/index.adoc
```diff
@@ -0,0 +1,124 @@
...
+It will follow a GitOps approach where SRE will request to revoke a certificate for a Kafka.
+Once the compromised certificate has been revoked, a new certificate for that Kafka will be generated.
+For already existing Kafkas, they'll continue to share the newly generated certificate: see proposal number (3). 
+
+6. When a standard Kafka is removed from the fleet manager, its corresponding certificate will also be revoked.
```
maleck13, line 72: same as above, is this effectively deleting the certificate ?

_adr/90/index.adoc
maleck13: cert-manager can work with multiple CertificateIssuers, you can set these issuers up via CRs. Then in the Certificate Resource you can decide which issuer you want to use. So another way you can mitigate is to have a multiple ACME providers and if one if failing request the certificate from another provider

_adr/90/index.adoc
```diff
@@ -0,0 +1,124 @@
...
+The proposed solution also limits the possibility of an owner of a enterprise data plane cluster, eavesdropping on another data plane cluster that they do not own as the certificate keys will be different. 
+
+NOTE: If the shared certificate for an existing Kafka is to be compromised, then it opens up a securiry risk on all existing Kafkas. However, the chances of this security breach occurring is mitigated by first and foremost ensuring only SREs have privileged access to data plane clusters. And second, if a the security issue is detected then certificates can be easily revoked and new ones autocreated 
+
+## Alternatives Considered / Rejected
+
+1. Deploying https://www.redhat.com/sysadmin/cert-manager-operator-openshift[cert-manager operator] to the data plane. Rejected to avoid exposing domain registrar credentials and to avoid adding another component that consumes data plane resources and that needs to be monitored.
+
+2. Migrating existing their Kafkas. This change is disruptive to existing customers hence why it was rejected.
```
maleck13, line 98: re-word?

_adr/90/index.adoc
```diff
@@ -0,0 +1,124 @@
...
+
+3. Deploying https://www.redhat.com/sysadmin/cert-manager-operator-openshift[cert-manager operator] in the Control Plane. 
+The Control plane then handles the pushing of the Certificate request CR, watching for the created secrets and storing them securely. 
+Rejected because this is technically complex and has no clear benefit over the propsed solution of using a library.
+It also violates the design intentions of the fleet manager by turning into an Kubernetes operator. 
+Finally, it's another component deployed that has to be monitored and managed in the control plane which forces every one (especially dev environments) to have a running Kubernetes cluster running during development
+
+## Challenges
+
+1. The drawback is that the solution can't or won't handle the migration of already existing Kafkas.
```
maleck13, line 109: while it is covered in the alternative. A challenge I see here is that we are effectively turning the fleet manager into a single ACME provider cert-manager and so are likely to run into bugs and issues that cert-manager has nailed down over time

machi1990: The long term is that HCG might act as cert broker, in the short term and to make it clear, the fleet manager doesn't implement this logic. We reuse existing ACME libraries - and leaving the cert management to the ACME library.

_adr/90/index.adoc
machi1990: The way that the flow work is that it'll actually call LE and revoke the certificate with `keyCompromised` reason.

machi1990:
The reviews feedback on this PR has been addressed. Can we get this one merged? Thanks @emmanuelbernard @tombentley

machi1990:
Thanks @tombentley

PR #86, opened 2023-01-25
ADR 90 OpenAPI clients code generation - Kiota

andreaTP:
Proposal to switch to a different generator technology for the MAS SDKs.


andreaTP:
Ironically superseeded by #90

Issue #87, opened 2023-01-26
OpenAPI clients code generation - Kiota


andreaTP:
Reference the [Original doc](https://docs.google.com/document/d/1WayfKKCvJn6gFmVRYbmBNTGixEF40vAMhdrlCDMLFuI/edit).

andreaTP:
cc. @tombentley please assign this issue to me.

tombentley:
@andreaTP it's an AP you want, right? (not an ADR)

andreaTP:
Hi @tombentley , thanks for taking a look!

I'm finding it a little confusing to identify if I should submit an AP or an ADR.

Currently in the available documentation: https://github.com/bf2fc6cc711aee1a0c2a/architecture/blob/main/pages/about/index.adoc#architectural-pattern

I can see that, in the `AP` description, the term `ADR` is used interchangeably and I'm finding it hard to identify the key differences. Would you mind clarifying or directly suggesting the best curse of action?

tombentley:
ADR records a decision (specific to a single service) that MUST be followed in applicable circumstances. The only way to avoid it is to supersede the decision.

PADR records a decision (applying to all services) that MUST be followed in applicable circumstances. The only way to avoid it is to supersede the decision.

Obviously these are a bit written-in-stone. So AP provides mechanism for saying "We think this is what we want to do, but we want to see whether the pattern is really generally applicable before we promote it to PADR". 

From the draft document you linked, I think this should be an AP. Once we've got experience with Kiota for at least one service we could have an PADR to record that it's what everyone MUST use from now on. 

Does that sound OK/work for you?

andreaTP:
Thanks a ton for the clarification.

Yes, I'm fine with submitting it as an AP!

andreaTP:
@tombentley the number given by the automation is colliding on @shawkins https://github.com/bf2fc6cc711aee1a0c2a/architecture/pull/82

tombentley:
Sigh. Avoiding this kind of thing is why the automation allocates the id in an initial (basically empty) PR. It's not the fault of @shawkins for not knowing that (because the bot is only now working thanks to @grdryn). 

@andreaTP I'll bump your AP id manually. That should be enough to fix it for future APs too.

tombentley:
@andreaTP you now have ADR 18 (you can re-fetch `main`).

andreaTP:
Cool, thanks for sorting things out! I'll continue to work on `AP 18`

PR #90, opened 2023-01-31
ADR91 / AP18 content

andreaTP:
Content of AP18.


andreaTP:
Hi @tombentley , how would I add reviewers to this AP?

_ap/18/index.adoc
```diff
@@ -3,76 +3,194 @@ num: 18
...
+In the long term, we expect that most of the bug fixes are going to happen in those shared libraries, and being able to maintain and release them at our own pace is a key element for success.
+
+==== Compatibility matrix
+
+Kiota https://github.com/microsoft/kiota/issues/2020[compatibility matrix] covers the current Red Hat use-cases, specifically:
+
+* Java
+* Go
+* Python
+* TypeScript
```
tarilabs, line 134:
```suggestion
* JavaScript
```

typo?

tarilabs: line just below is TypeScript, so TypeScript was listed twice

andreaTP: thanks for the suggestion!
This is actually deliberate, even the current "js-sdk" is producing TS:
https://github.com/redhat-developer/app-services-sdk-js

_ap/18/index.adoc
```diff
@@ -3,76 +3,194 @@ num: 18
...
+Transitive dependencies are kept up to date in a timely manner.
+
+Improvements and bugs are continuously integrated into both the generator and the auxiliary libraries' repositories.
+
+==== Swappable implementation libraries
+
+The design of Kiota allows consumers of the code generator to provide their own https://microsoft.github.io/kiota/extending/corelibrary.html[“opinionated” implementations] of core generated client functionality.
+This is done via a set of interfaces to core functionality that all generated clients leverage (e.g. a Java HTTP client interface).  Default implementations of these interfaces are provided, but alternatives can easily be created and used.
+
+Specifically for the Red Hat use case, we are free to roll out our own implementations of the core libraries to better support Red Hat products and strategy (e.g. Quarkus specific implementations of certain functionality).
```
tombentley, line 120: It would be good to spell out the implementations we're expecting to write. HTTP client is one, but what about authentication, for example?

tombentley (approved):
This seems reasonable to me, though I'm not an expert in this area.

_ap/18/index.adoc
andreaTP: Good question, regarding implementations, for Java I'd expect:
- a Vert-X HTTP client implementation (as opposed to the default OkHttp)
- a JSON serialization based on Jackson (as opposed to the default Gson)

Do we have strong opinions on the other languages @tombentley ?

@riccardo-forina can you check the TS libraries? https://github.com/microsoft/kiota-typescript
@dimakis can you have a look over Python? https://github.com/microsoft?q=kiota+python&type=all&language=&sort=
@lburgazzoli can I ask for your support in Go? https://github.com/microsoft?q=kiota+go&type=all&language=&sort=

🙏  thanks in advance!

Regarding Authentication I think that we should support only OFFLINE_TOKEN, this means rolling out a custom `TokenProvider` for each language (it's not mandatory but highly conveninet). this is in Java for example:
https://github.com/andreaTP/kiota-utils#redhat-sso-auth

andreaTP: @tombentley addressed here: https://github.com/bf2fc6cc711aee1a0c2a/architecture/pull/90/commits/c9f2cee1fc624782620d4ca59498b5bf64c50ca1 let me know if you prefer additional detail.

andreaTP:
@tombentley anything I can do to speed up this process?

I start to feel uncomfortable having a PR open for > 1 month where ppl are expected to provide additional feedback.

lburgazzoli:
LGTM

_ap/18/index.adoc
```diff
@@ -3,76 +3,198 @@ num: 18
...
+
+The DevExp team has identified as a major piece of maintenance the upgrade of the currently outdated https://github.com/OpenAPITools/openapi-generator[openapi-generator].
+The current SDKs are full of quirks and edge cases in most cases hacked around to end up with a working implementation.
+Having to keep the https://docs.google.com/spreadsheets/d/1fwmRmapDdFRU-B7qFBUuXyHMDjzXtRMY-jhbtxDVtyU/edit#gid=0[hacks and workarounds] is a major blocker to updating toward newer versions of the generator.
+
+The monolithic approach to a generation used by openapi-generator also offers challenges in a modular approach to bug fixes and improvements.
+
+==== IBM
+
+We have had extensive discussions with IBM regarding their approach to https://github.com/IBM/ibm-cloud-sdk-common[client SDK] generation and we very much like their governance model over APIs produced in the company.
```
maleck13, line 184: I am missing why this was ruled out?

maleck13 (approved):
LGTM just one comment on clarifying why the IBM tool was ruled out

_ap/18/index.adoc
andreaTP: The section on IBM was based on confidential information.
I stripped it down to a minimum that can be safely shared.

Which are the options to keep part of this document private?
Should I make an "Addendum" in the form of a Google Doc linked here?
@tombentley do you have any suggestion here?

andreaTP:
Heads up to @danielezonca @emmanuelbernard @EricWittmann for a review 🙏 

Please, feel free to ping me in chat if I can do anything to facilitate it.

_ap/18/index.adoc
tombentley: GDoc seems like a good idea to me @andreaTP

andreaTP: done: https://github.com/bf2fc6cc711aee1a0c2a/architecture/pull/90/commits/6bf32c97b56f4c73c3b363481f23ef38b8e854b6

EricWittmann (approved):
lgtm

_ap/18/index.adoc
```diff
@@ -3,76 +3,200 @@ num: 18
...
+
+Specifically for the Red Hat use case, we are free to roll out our own implementations of the core libraries to better support Red Hat products and strategy (e.g. Quarkus specific implementations of certain functionality).
+
+Creating alternatives for the provided default libraries is doable in a very reasonable amount of time.
+https://github.com/andreaTP/kiota-utils/blob/1a7dcecd92ce8794fe479e9c37c71d5733a39ff6/http/jdk-http/src/main/java/io/apicurio/kiota/http/JdkHttpRequestAdapter.java#L1[Here] you can find an alternative implementation of the Java HttpClient interface that is based on the client embedded in JDK 11+ (vs. the out-of-the-box implementation that is based on OkHttp).
+We expect to provide:
+
+- one `RHTokenProvider` that will work automatically with [offline tokens](https://access.redhat.com/articles/3626371) and SSO like [we did for Java](https://github.com/andreaTP/kiota-utils/blob/93083650e0374f511abefd099033f05276927deb/rh-auth/src/main/java/com/github/andreatp/kiota/auth/RHAccessTokenProvider.java#L24) for each supported language.
```
danielezonca, line 126: In the preview these links seems to be malformed (probably you are using markdown syntax instead of adoc)
```suggestion
- one `RHTokenProvider` that will work automatically with https://access.redhat.com/articles/3626371[offline tokens] and SSO like https://github.com/andreaTP/kiota-utils/blob/93083650e0374f511abefd099033f05276927deb/rh-auth/src/main/java/com/github/andreatp/kiota/auth/RHAccessTokenProvider.java#L24[we did for Java] for each supported language.
```

_ap/18/index.adoc
```diff
@@ -3,76 +3,200 @@ num: 18
...
+* Clients with large numbers of dependencies could be vulnerable if not refreshed
+* A swapped implementation may not be well supported
+
+=== Alternatives Considered / Rejected
+
+==== Keep going with the current generator
+
+The DevExp team has identified as a major piece of maintenance the upgrade of the currently outdated https://github.com/OpenAPITools/openapi-generator[openapi-generator].
+The current SDKs are full of quirks and edge cases in most cases hacked around to end up with a working implementation.
+Having to keep the https://docs.google.com/spreadsheets/d/1fwmRmapDdFRU-B7qFBUuXyHMDjzXtRMY-jhbtxDVtyU/edit#gid=0[hacks and workarounds] is a major blocker to updating toward newer versions of the generator.
```
danielezonca, line 178: This link is not publicly accessible, I don't know if it makes sense to link it here

_ap/18/index.adoc
```diff
@@ -3,76 +3,200 @@ num: 18
...
+* Better developer experience
+* Maintenance
+* Swappable implementation libraries
+* Compatibility matrix
+
+=== Disadvantages:
+
+* Not GA (yet)
+* External Owners (Microsoft)
+* Core Generator codebase is .NET
```
danielezonca, line 87: It is not a key aspect for this ADR but does it imply that this library cannot be integrated with Quarkus to replace the existing Quarkus OpenAPI extension or in general any possible Quarkus similar use cases?

danielezonca (approved):
Few minor comments but overall LGTM, very well done @andreaTP 👍

_ap/18/index.adoc
andreaTP: thanks!

_ap/18/index.adoc
andreaTP: Let me try to provide more context here:

- `this library cannot be integrated with Quarkus` -> Kiota ships as a standalone CLI + Java dependencies and will work with Quarkus (possibly it will need a little bit of tweaking for `native-image`, but it should be it). We are already publishing a Maven plugin to easily integrate with Java workflows: https://github.com/redhat-developer/kiota-java-extra#maven-plugin
- `replace the existing Quarkus OpenAPI extension` -> replacing the current extension is out of the scope of this AP and out of the scope of Kiota. The Quarkus extension is vertical and highly integrated with an opinionated language and framework, while Kiota privileges a more consistent approach across different languages and decouples over the used frameworks. I plan to agree on an action plan with Ricardo to see if it makes sense to support both approaches, but no strong commitment here.
- `or in general any possible Quarkus similar use cases?` -> can you, please, elaborate if there is any aspect that I might have missed with the previous answers?

andreaTP:
@emmanuelbernard can you please have a look at this? 🙏  alternatively can you nominate a delegate so that we can move forward?

Thanks in advance!

_ap/18/index.adoc
danielezonca: Is the content of this link private? If so, fine for me to follow the same approach :)

_ap/18/index.adoc
danielezonca: Thanks for the clarification, you covered all my questions :)

_ap/18/index.adoc
andreaTP: ouch, my bad, I haven't noticed this is actually a different link, I think we can safely drop it 👍

andreaTP: removed

_ap/18/index.adoc
```diff
@@ -3,76 +3,200 @@ num: 18
...
+
+It's also pretty common to have breaking changes in the generator tooling, which results in breaking changes in the generated code.
+When this happens it can be hard to detect and often happens without a clear migration path.
+
+For these reasons we are often https://github.com/redhat-developer/app-services-sdk-go/blob/13d1cad9a937c456a9623ad188c44790e3370d83/openapitools.json#L5[stuck on old versions], preventing (in some cases) the opportunity to fix CVEs and issues by keeping transitive dependencies and used frameworks updated.
+
+Following are a variety of critical/important issues we currently have due to our OpenAPI Generator based approach to API Client generation.
+
+=== Security
+
```
emmanuelbernard, line 45: Discuss mitigation on security issues coming from. Probably iso risk between old and new approach but make it clear of the security risk paths.

_ap/18/index.adoc
```diff
@@ -3,76 +3,200 @@ num: 18
...
+* Dictate a technical decision for internal usage or for projects that are successfully rolling their own SDKs and clients (e.g. by manually curating them)
+
+== Current Architecture
+
+Inside Red Hat we are mostly relying (in various ways) on OpenAPI Generator to generate clients, we also maintain a https://github.com/quarkiverse/quarkus-openapi-generator[Quarkus Extension] based on the same generator.
+
+== Proposed Architecture
+
+We propose to switch to alternative OSS tooling provided by Microsoft called https://github.com/microsoft/kiota[Kiota].
```
emmanuelbernard, line 72:
```suggestion
We propose to switch to an alternative OSS tooling provided by Microsoft called https://github.com/microsoft/kiota[Kiota].
```

_ap/18/index.adoc
emmanuelbernard: Ah it's in the Thread Model section

emmanuelbernard:
I like the decision but here it feels like this is an ADR level decision, not an AP level one.
I'd capture this content as ADR for the effective migration.
And write a shorter future usage focused doc that explains the what services should/could use for their API generation. (less focused on the community migration).
Thoughts?

andreaTP:
@emmanuelbernard Thanks a lot for reviewing, appreciated!

> I like the decision but here it feels like this is an `ADR` level decision, not an `AP` level one.

This proposal was initially submitted as an ADR and got converted into an AP after this comment:
https://github.com/bf2fc6cc711aee1a0c2a/architecture/issues/87#issuecomment-1408771243

I'm personally not fussy about this aspect and am open to suggestions (this is, anyhow, my first proposal to the architecture group and inexperience might show).
If you want me to convert this back into an ADR, that's completely fine for me as soon as @tombentley gives a nudge.

On a side note, I'm sorry to call this out, but I'm afraid to notice that everyone is constantly confusing and mixing up `AP` with `ADR`, but I'll file a separate issue to have this tangential discussion.

emmanuelbernard:
Let's have a quick chat about it tomorrow at the arch call to clear things up.

tombentley:
On the point of AP vs ADR... FTR based on the original model we had in mind:

* If you were recording the decision to do this for a single service (e.g. RHOSAK) then it would be an ADR. 
* As I understand it, while this proposal uses RHOSAK as an example, you're actually confident enough to be *recommending* this for other services. I.e. BFF should be the default way of solving this problem, and any service wanting to deviate ought record those reasons in an ADR for that service.
* Once the pattern has been proven in several services (nominally 3) you might write a PADR to record the decision that all services SHOULD use this pattern.

The motivation for AP was to not be too quick to generalise the experience of one service (ADR) into a hard and fast rule for all other services to follow (PADR). 

So to me we're wanting to extract an AP here. There's two ways we could do that:

1. Write an ADR for RHOSAK (for which the level of detail you have here is about right) and then summarise that in an ½ page AP which basically says "do like in this ADR for other services".
2. Just write the AP (which is what you've done in this case). A bit more detail is appropriate in this case since there isn't a more detailed ADR to point to.

andreaTP:
Thanks for the write-up @tombentley !
I'll try to wrap up as proposed and we move on from there.

_ap/18/index.adoc
```diff
@@ -3,76 +3,49 @@ num: 18
...
-// useful structure for writing a clear document.
-
 ## Intent
-// Summarize in a single sentence what the pattern tries to achieve.
-//
-// For example:
-//
-// Define a policy for exposing metrics to customers in a consistent way.
+
+Define a consistent technology to be used for client code generation from OpenAPI specifications.
```
emmanuelbernard, line 12:
```suggestion
Describe a path to generate client code from OpenAPI specification for our service APIs.
Lower the burden of code generation from OpenAPI specifications
```

_ap/18/index.adoc
```diff
@@ -3,76 +3,49 @@ num: 18
...
 
 ## Applicability
 
-// Call out when this pattern might apply (and when it should not apply, if relevant)
-// For example:
-//
-// This pattern applies to any service which exposes metrics as Prometheus-scrape endpoints.
-// It does not apply to services which provide access to metrics in other formats, or via
-// means other than an API (e.g. metric visualization via a system such as Grafana).
+This pattern applies to any service using a client/SDK to access REST API endpoints.
```
emmanuelbernard, line 22:
```suggestion
This pattern applies to any service providing a client/SDK to access REST API endpoints.
```

_ap/18/index.adoc
```diff
@@ -3,76 +3,49 @@ num: 18
...
+
+To apply Kiota in a project follow the https://microsoft.github.io/kiota/get-started/[Official "Get Started" guides].
 
 ## Participants
 
-// What components and/or teams are directly involved
-//
-// For this metrics example this would include the customer and the control plane via which
-// the metrics are exposed
+The DevExp team is currently providing SDKs for all of the MAS services, all of them will be converted to use Kiota and serve as an example that can be applied more broadly.
```
emmanuelbernard, line 36:
```suggestion
The DevExp team is currently providing SDKs for all of the bf2 services, all of them will be converted to use Kiota and serve as an example that can be applied more broadly.
```

_ap/18/index.adoc
```diff
@@ -3,76 +3,49 @@ num: 18
...
-// For example:
-// 
-// Prometheus scrape endpoints are the defacto standard for exposing metric information for collection.
-// However this format and the associated conventions are not sufficient in themselves to ensure
-// that metrics are exposed in a consistent way. Problems include:
-// * Insufficient guidance on metric and label naming to provide consistency and 
-//   establishing the principal of hiding implementation details.
-// * Treatment of this as a first-class API, with established mechanisms for API evolution
-// * etc.
+* Have a uniform and good quality SDK layer to be used for APIs access
```
emmanuelbernard, line 16: * offer a client or SDK for Services APIs that CLI, UI and customers can rely on

_adr/91/index.adoc
```diff
@@ -0,0 +1,204 @@
+---
+num: 91
+title: OpenAPI clients code generation - Kiota
```
emmanuelbernard, line 3:
```suggestion
title: Generating clients code from OpenAPI contracts for our service APIs
```

_ap/18/index.adoc
andreaTP: this is actually not correct, Kiota is going to be used on the "receiver side", and possibly even without the need of curated SDKs

_ap/18/index.adoc
andreaTP: The fact that I left this one out was actually intentional ... I can get it back, but my intention was to:

> leave the door open to (eventually)provide only "tooling" and not full-fledged SDKs

I understand this might sound a little excessive at this stage but, at least, I would like to keep this door open.
Thoughts?

emmanuelbernard: When you leave the door open, you assume a lot of imagination from a casual reader looking to know if his problem is solved. And that reader ahs to read a lot of things so won't meditate on it. I would go for explicit, you can add tooling as an extra option but the main use case being the sentence you left out.

_ap/18/index.adoc
emmanuelbernard: So the intent is not to offer SDKs (generated). But to use Kiota direcly in the CLI and UI build system?

_ap/18/index.adoc
andreaTP: ok, fair 👍

_ap/18/index.adoc
andreaTP: integrated with `using or providing`

_ap/18/index.adoc
andreaTP: added

andreaTP:
Linked AP18 to ADR91 so people can easily find the long story.
Ready to merge? We have all the needed green checks now as far as I can tell! 🎉

tombentley:
Thanks @andreaTP

PR #97, opened 2023-03-10
Improve Architecture Pattern template to help writers make the most useful document

emmanuelbernard:


tombentley (approved):
Suggested a few rephrasings (and applied "one sentence per line as a meta point).

PR #99, opened 2023-03-14
ADR-92 & AP-19: BFF proposal

andreaTP:
Opening this PR as Draft to foster collaboration on the Proposal itself and for early comments.

cc. @EricWittmann @riccardo-forina @MikeEdgar @dimakis @lburgazzoli


andreaTP:
Question for @tombentley and @emmanuelbernard , I got again confused regarding AP vs. ADR 😢 
@tombentley booked it as AP but:

- it doesn't automatically apply to specific situations and highly depends on the current system setup and capabilities
- it's not a generic solution or pattern as it highly depends on the design(accidental or deliberate) of the rest of the system
- I do not want to "sell BFF" as a general solution as, almost everything here, is based on tradeoffs

I ended up framing it as an ADR, let me know if this works for you.

emmanuelbernard:
If you feel BFF should not be used for other services like RHOC or whatever is coming up, then it is the right choice. I had a different impression in our conversation last Wed.

andreaTP:
> If you feel BFF should not be used for other services like RHOC or whatever is coming up, then it is the right choice. I had a different impression in our conversation last Wed.

I haven't analyzed the current situation in RHOC to be able to answer your question 😢 , let me gather @lburgazzoli feedback to address this.
If we reach an agreement I'll bundle this ADR up with an AP.

emmanuelbernard:
> If we reach an agreement I'll bundle this ADR up with an AP.

Not bundle, since AP have a different intent and template, it's a partial rework.

_adr/92/index.adoc
```diff
@@ -0,0 +1,102 @@
...
+  - Dev Exp
+applies_padrs: # What PADRs does this ADR apply?
+applies_patterns: # What APs does this ADR apply?
+---
+
+## Context and Problem Statement
+
+The integration of the Enterprise/Hybrid option of RHOSAK in UI, SDKs and CLI has shown that there is a margin for improvement in how we expose and consume our services APIs.
+
+* the current services are offering the bare functionality leaving room for integration code to live in the services' consumer codebases
```
EricWittmann, line 21: I'm not sure I understand this 100%.  Can you explain a bit more what you mean and maybe we can reword a bit to be more clear?

_adr/92/index.adoc
```diff
@@ -0,0 +1,102 @@
...
+* the data aggregation logic is duplicated, scattered, and eventually diverging in various codebases (especially UI and CLI)
+* there is no "single source of truth" for data aggregated from various services
+* the current REST API layer exposes the user to the entire complexity of the underlying infrastructure
```
EricWittmann, line 24: Is it the complexity of the infrastructure that's exposed or is it the complexity of the main service PLUS the supporting services (e.g. AMS)?  Infrastructure makes me think we're exposing network routing and whatnot.  That may be true for RHOSAK Enterprise maybe?

_adr/92/index.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+We want to deduplicate the development effort and provide a more cohesive UX to the users.
```
EricWittmann, line 26: This point is critical.  Those are the two main goals as I see it:

1. Single source of logic shared by UI and CLI
2. Better API tuned specifically to common use-cases

EricWittmann: Ignore this - I didn't read ahead to the Goals section.  :)  +1 to the goals.

EricWittmann:
Added some comments but looks good overall.  Something that's missing perhaps is a comment about potential lack of control (not now I guess but maybe in the future) over APIs being aggregated.  Another benefit of the BFF approach is when we are running a service using e.g. an Apache project, we might not be able to modify the API to make it more suitable for use by the UI/CLI (something you could argue we **can** mostly do today).

_adr/92/index.adoc
andreaTP: Tentatively:
```
the services are exposing APIs for the offered specific functionality, but we lack an integration layer
```

what we are trying to say is that there is no integration logic exposed by the current services.
The services are offering APIs modeled after their specific functionality and are very vertical, but any operation that involves more than one service becomes difficult and involves writing the integration logic on the client side.

Feel free to propose a better working 🙏 !

_adr/92/index.adoc
andreaTP: Renaming `infrastructure` -> `architecture`.
We do not expose (much)"infrastructure" detail, but the user really needs to know the architecture of the system to navigate it.

lburgazzoli:
> > If you feel BFF should not be used for other services like RHOC or whatever is coming up, then it is the right choice. I had a different impression in our conversation last Wed.
> 
> I haven't analyzed the current situation in RHOC to be able to answer your question cry , let me gather @lburgazzoli feedback to address this. If we reach an agreement I'll bundle this ADR up with an AP.

For RHOC things are exacerbates by the fact that we must interact with multiple services, i.e. Kafka & Service Registry so we would benefit from a BFF that knows how to interact with such services (i.e. to set-up/check permissions, gather additional information, etc).

andreaTP:
Added the AP proposal, marking as Ready For Review now

riccardo-forina:
Could we also try to solve the `subscription` problem as well? We use the polling pattern now to get the data we need when it changes, but it's not the cleanest solution as it unnecessarily requires a lot of micromanagement and hammers the backend.

_adr/92/index.adoc
```diff
@@ -0,0 +1,104 @@
...
+### Expand the current services APIs
+
+We want the current microservices to stay vertical and opinionated toward a minimal and curated API that directly maps the functionality of the service without having to deal with the integration challenges.
+
+## Challenges
+
+* Caching while guaranteeing multitenancy
+* Shared ownership of the component across different teams
+* Avoid API proliferation after the first BFF is in place
+* Potential lack of control over APIs being aggregated
```
MikeEdgar, line 94: Another challenge worth noting is that with the notion of private data planes the control plane may not have access to the back-end APIs or data sources.

gashcrumb:
I'd suggest just a BFF isn't enough.  I think offloading client-specific concerns such as handling update subscriptions and caching is great and we should adopt a BFF for those reasons to support each client-side experience.

But I'd also suggest we probably need a separate business integration service that handles the orchestration of the service APIs and needs to be adaptable to changing business concerns, kind of the two things we currently try and spread between our UI and API a bit.

_adr/92/index.adoc
andreaTP: added

_adr/92/index.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+### Threat model
+
+Being it a publicly accessible service the generic service Threat model applies, we should ensure stability of the platform by having all the good practices in place:
+
+* Rate Limiting
+* Circuit Breaking
+* Caching
+
+Regarding multitenancy, it can be achieved, in the first prototype by re-using the Access Token for the initial API request to issue the following requests to the backend services.
```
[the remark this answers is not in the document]
alechenninger: In general this is something we want to facilitate, though sometimes the right thing to do is make a new request with a client credential grant (i.e. a service account authorization grant as opposed to the user's grant). It depends whether the request is really on behalf of the user, or whether it needs to be done with the identity of the BFF layer itself. This is a more complex issue of semantics and authorization which I'll leave out of this reply.

As far as making a subsequent request on behalf of the user, with a single SSO (a single issuer), yes, this becomes easier. But, a compatible issuer is not the only requirement. You also need to make sure the access token has the right audience and scope for downstream APIs. For audience ("aud" claim), we're simplifying this for new clients to be simply "redhat.com" for any hosted redhat.com APIs. Then, each API (aka "resource server) needs to define at least one scope (and ideally several) and enforce it. Then, the clients (frontend apps I believe in this case), need to be sure to request those scopes during their grants. On the SSO side, we can ensure that scope inclusions are transitive, so the client doesn't need to know about downstream calls. That is, the client might request a scope for the BFF to do something, and if that needs another service, which needs another service, those downstream scopes can be included automatically by SSO if previously negotiated between SSO provider (us) and those API providers when configuring the scopes.

See https://docs.google.com/document/d/1fv2LNl81avyimUGoHiVJzAKX3I-gZRu9cpiS0qDmz9Y/edit#heading=h.ds8q4xtkmu64 for more info.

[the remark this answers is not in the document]
alechenninger: One minor issue to consider with this, is if we ever get to using [sender-constrained tokens](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-dpop) or [signed HTTP requests](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-message-signatures), it becomes a bit more difficult to forward tokens along. But, this is probably worth writing off as a known challenge for future "us" to worry about ;-).

_adr/92/index.adoc
```diff
@@ -0,0 +1,105 @@
+---
+num: 91
+title: "Toward a more unified UX: BFF"
```
apupier, line 3:
```suggestion
title: "Toward a more unified UX: BFF (Back For Frontends)"
```

_adr/92/index.adoc
```diff
@@ -0,0 +1,105 @@
...
+Regarding multitenancy, it can be achieved, in the first prototype by re-using the Access Token for the initial API request to issue the following requests to the backend services.
+
+## Alternatives Considered / Rejected
+
+### GraphQL
+
+Is a well known technical platform for implementing BFFs.
+
+* Lack of expertise
+* We don't want a generic API
```
chirino, line 81: Why?

_adr/92/index.adoc
```diff
@@ -0,0 +1,105 @@
...
+* Challenges in caching
```
chirino, line 82: Most API calls typically are not cached..  Even then there are client side GaphQL clients that will cache queries for you.  I would remove this point.

_adr/92/index.adoc
```diff
@@ -0,0 +1,105 @@
...
+* Poor support for "commands"(as opposed to queries)
```
chirino, line 83: GraphQL mutations work just as good as queries.   I would remove this line too.

_adr/92/index.adoc
```diff
@@ -0,0 +1,105 @@
...
+* Rate Limiting
+* Circuit Breaking
+* Caching
+
+Regarding multitenancy, it can be achieved, in the first prototype by re-using the Access Token for the initial API request to issue the following requests to the backend services.
+
+## Alternatives Considered / Rejected
+
+### GraphQL
+
```
chirino, line 77: Given GraphQL is the best known standardized architecture for BFF's, I think it's rejection deserves a well thought out and reasoned argument.

chirino (changes requested):

_adr/92/index.adoc
andreaTP: > Given GraphQL is the best known standardized architecture for BFF's

Curious to look at the numbers/sources, can you please share?

> I think it's rejection deserves a well thought out and reasoned argument.

I'll review this section by the EOD with new input from a chat with @machi1990 and @EricWittmann

chirino: Don't have any numbers off hand.  Maybe someone can go to this years GraphQL conference and gather intel.  Or maybe find the info in some of last year's sessions: https://www.apollographql.com/events/virtual-event/graphql-summit/#on-demand

_adr/92/index.adoc
chirino: GraphQL fits the "where, possibly, one endpoint is enough to render a page/offer a functionality." requirement.  I don't see why you make the "* We don't want a generic API"  point against GraphQL then.

andreaTP: Deal, I'll remove this point 👍

_adr/92/index.adoc
andreaTP: Honestly interested in a more "independent" data source other than a conference on the technology!

E.g. I love the Scala programming language but I won't go to a Scala conference to gather data about "how much it spreads", there is an obvious bias there.

chirino: It is a great way to judge community and learn from other folks implementing it.  If the conference is small, then not to many folks use it or are interested in it.  Like what ever we come up with would have ZERO ecosystem in comparison.

_adr/92/index.adoc
andreaTP: done

_adr/92/index.adoc
andreaTP: done

_adr/92/index.adoc
```diff
@@ -0,0 +1,107 @@
...
+Is a well known technical platform for implementing BFFs, especially strong for composing data from multiple services but it doesn't magically solves the problem by itself.
+When comparing a possible OpenAPI solution to a GraphQL one, for this specific problem, the real advantage in favor of the latter is the possibility for dynamic fields selections in the client.
+We do not consider this upside strong enough to outweighs some of the challenges:
+
+* Lack of expertise in the technology - one of the objective of this component is to enable multiple teams collaboration on the repo
+* Challenges with Observability - Support for partial failures compromise the usage of "generic" REST APIs monitoring tools
+
+We leave the doors open to re-evaluate the choice of GraphQL in the future and/or in different contexts.
```
machi1990, line 85: +1 if needed, extending/offering GraphQL API on top of what's proposed here is possible e.g via https://quarkus.io/guides/smallrye-graphql

The datasources (several underlying core APIs) and how they are composed to return a certain `View/Representation` will remain the same logic in most cases - and just exposed differently.

EricWittmann (approved):
LGTM

_adr/92/index.adoc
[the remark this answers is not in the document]
emmanuelbernard: In the context of BFF, the capture is that we should prefer that we can propagate the access token throguh to the dependent services because it will keep the development of BFF easier and keep the security model similar to a usage without BFF.

_ap/19/index.adoc
```diff
@@ -1,78 +1,58 @@
...
+In those cases placing a frontend opinionated component in front can be an appropriate solution to speed up development and be a single source of truth for integration logic.
 
 ## Description
 
-// Describe the pattern in enough detail that a fresh pair of eyes can understand it
-// and apply it elsewhere.
-//
-// For this metrics example we might link to existing guidelines on naming, augment them with extra rules 
-// which apply to services, explain how the API would allow for graceful API evolution without "flag days".
+Implementing a https://samnewman.io/patterns/architectural/bff/[BFF] for the backend services to better decouple and centralize the integration logic.
```
emmanuelbernard, line 31: Add the `propagate the access token` concept in the description.
Mention the alternative being GraphQL but that for the team, expertise was not deems high enough to explore that path. This will leave the doors for other teams to explore.

andreaTP: 1. done
2. I do believe that the current wording reflect weel enough our position toward GraphQL: https://github.com/bf2fc6cc711aee1a0c2a/architecture/pull/99/commits/877e4c8dac5be9ea3b2736bb0c73004adea04531

andreaTP:
A polite ping to @danielezonca @maleck13 @tombentley @lburgazzoli for the review! Thanks in advance 🙏

lburgazzoli (approved):

_adr/92/index.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+We want the current microservices to stay vertical and opinionated toward a minimal and curated API that directly maps the functionality of the service without having to deal with the integration challenges.
+
+## Challenges
+
+* Caching while guaranteeing multitenancy
+* Shared ownership of the component across different teams
+* Avoid API proliferation after the first BFF is in place
+* Potential lack of control over APIs being aggregated
+* The introduction of private data planes can introduce additional issues because of backend APIs reachability
```
maleck13, line 98: This would seem to be a real problem. Not supporting private clusters seems like quite the architectural draw back given we want to address the enterprise use case?

maleck13 (approved):
A concern I have here is called out as a challenge, it does not work for private data planes, which could restrict the potential customer / user base and whether that is something that is an acceptable tradeoff.

_adr/92/index.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+We leave the doors open to re-evaluate the choice of GraphQL in the future and/or in different contexts.
+
+### Expand the current services APIs
+
+We want the current microservices to stay vertical and opinionated toward a minimal and curated API that directly maps the functionality of the service without having to deal with the integration challenges.
+
+## Challenges
+
+* Caching while guaranteeing multitenancy
```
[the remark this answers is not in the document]
riccardo-forina: The old APIs are staying, so it's not a challenge. Or at least, it's not any different than today.

_adr/92/index.adoc
riccardo-forina: I think the bulk of the issue relies on the control plane, which requires clients to connect and manipulate many responses from different services. We bleed a lot of architectural decisions and problems to clients.

The data plane is easier; clients connect with the instance and work with the instance directly. There is less need to connect with other services (as of today), so I think we can keep clients connecting with the data plane directly.

Still, the lessons learned can and should be applied to the data plane for future developments.

andreaTP:
> it does not work for private data planes

@maleck13 thanks a lot for the review! I do believe that we have full freedom to move this component to the data plane when/if necessary.
We marked it as an open point as it depends on other architectural changes and proposals that are being discussed those days.
Does this and Riccardo's answer your question?

_ap/19/index.adoc
emmanuelbernard: On 2, it's in the ADR and not in the AP, It is worth repeating to help consumption of the materials.

andreaTP: Any reference to the implementing technology has been deliberately omitted from the AP, unless you have a really strong opinion I would keep it this way.

emmanuelbernard: Yes I think it's important. It's not an implementation detail to go for a custom REST API vs a GraphQL based API, this is a defining choice.

andreaTP: > this is a defining choice

can you please elaborate?
This interpretation is completely alien to me and, I believe, to anyone else involved in drafting this proposal.
cc. @EricWittmann @riccardo-forina

danielezonca (approved):
A minor comment but LGTM 👍

_adr/92/index.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+image::current_architecture.png[Current Architecture]
+
+
+## Proposal
+
+[NOTE]
+*Assumption* as a first iteration we are assuming that the CLI needs can be covered by a strict subset of the functionalities needed for the UI.
+
+We propose to integrate *one* server side component commonly known as https://samnewman.io/patterns/architectural/bff/[BFF] (Backend For Frontend).
```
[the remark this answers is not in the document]
andreaTP: > so this is an additional endpoint is going to be part of the API delivered by that specific service.

this is not correct, here we are proposing BFF as a separate service with a new spec that is going to aggregate multiple different services.
We are somehow covering only what you call "cross service BFF", as having one BFF associated with one service almost defeats the purpose of BFF (at least in this context).
Also, we are here proposing a single BFF service for the entire platform.

If this is not clear, please, point out the confusing sections and I will try to re-phrase 👍

_ap/19/index.adoc
riccardo-forina: I'm also for not mentioning any particular technology here. A BFF could be implemented in a number of ways and there is no "best" or "worst" technology, but a "right" one that fully depends on the context.

For example, for a team that's using Typescript for the full stack and fully control the clients connecting to the backend, tRPC is a great choice, arguably better than GraphQL or plain old REST+OpenAPI

_adr/92/index.adoc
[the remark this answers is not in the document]
danielezonca: Thanks for the clarification, fine for me to keep current phrasing 👍

_ap/19/index.adoc
andreaTP: @emmanuelbernard if you are convinced that this aspect is important for the AP I'm happy to remove the AP from this PR and let you(or someone else) take over in another iteration.
In this case, I disagree with mixing scopes:
- BFF is the Architectural Pattern we want to promote as a possible solution to common issues
- OpenAPI/GraphQL is the technical instrument used to implement it, which is a completely orthogonal discussion - i.e. it can be implemented with CORBA or any other technology that serves the purpose

We briefly touched on this point with @chirino and I felt alignment on this aspect.

andreaTP:
@tombentley can you please, review or delegate the review of this PR? Thanks in advance 🙏

_adr/92/index.adoc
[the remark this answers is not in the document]
chirino: > here proposing a single BFF service for the entire platform.

Define entire platform.

_ap/19/index.adoc
chirino: Lol you mention COBRA in the same breath as GraphQL as being fitting for a BFF pattern.  I’m not sure you understood our conversation at all.

_adr/92/index.adoc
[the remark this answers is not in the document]
emmanuelbernard: I added the detail in a subsequent commit.

_adr/92/index.adoc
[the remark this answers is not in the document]
emmanuelbernard: Entire platform means RHOSAK RHODS RHOC  as I understand it

_ap/19/index.adoc
emmanuelbernard: I am proposing changes as part of a new commit. See https://github.com/andreaTP/architecture/pull/1

emmanuelbernard:
Hey @andreaTP @riccardo-forina and all,

In order to help move forward, I've sent a PR to your PR (not sure how well it will work).
https://github.com/andreaTP/architecture/pull/1

I've closed some conversations in the ADR that were pending here.
I've also adjusted the AP to show what I was trying to convey and the choices that I think are important.
In particular, I want it to be closer to actionable than theoretical.

emmanuelbernard:
I tried to add commits to Andrea's own branch but without success, hence the PR

_adr/92/index.adoc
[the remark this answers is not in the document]
danielezonca: I don't think we have enough context to include RHODS in the scope of this decision, I would "limit" the scope this ADR to RHOSAK+RHOC+RHOSR and postpone to a revision RHODS but also RHOAM and other services

tombentley:
@andreaTP this was top of my todo list today, but do you want me to review this now, or once you've merged Emmanuel's PR into the branch for this PR?

_adr/92/index.adoc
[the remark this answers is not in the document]
emmanuelbernard: From what I foresee, we won't have a choice if we want a good UX. So let's go with this bold intent and cast exclusions as we go.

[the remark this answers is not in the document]
danielezonca: I see this as a very _bold_ intent given that RHODS for example has the on-prem as a very important setup and also private cloud data plane configurations are not fully explored. :)

Btw I was mainly referring to the ADR that is "Kafka ecosystem" specific while the AP is probably generic enough and I agree that we aim to have a coherent UX across the whole portfolio so the problem statement and approach for the solution can be valid for every service

andreaTP:
thanks @emmanuelbernard for the PR, merged 👍 
@tombentley ready for review now

_ap/19/index.adoc
andreaTP: Thanks @emmanuelbernard , the context you proposed is fine for me, merged 👍 

@chirino I firmly believe that CORBA is a perfectly valid fit for an application in COBOL ( [ref](http://computer-programming-forum.com/48-cobol/f4fdc89fc2d9bfb3.htm) )

_ap/19/index.adoc
```diff
@@ -1,78 +1,78 @@
...
+image::with_BFF.png[Proposed Architecture]
+
+Following are key contextual information to understand the pattern and its potential limitations.
+
+Access to the underlying services will be done by propagating the original (user) access token to simplify development and to keep the security model similar to the current architecture.
+See ADR 91 for more details.
+
+The implementation choice went for a custom REST endpoint implementation as opposed to using GraphQL.
+GraphQL is a popular technology built to address this or a similar set of use cases.
+The team's experience lead tot he choice of a custom REST endpoint but could be revisited by this or another team.
```
tombentley, line 47:
```suggestion
The team's experience led to the choice of a custom REST endpoint but could be revisited by this or another team.
```

andreaTP:
Ready to merge AFAIU 🙏 thanks everyone for the input!

